PR #1, opened 2018-11-02
First pass at UI canvas

mklewandowski:
More work needs to be done here, but this is a starting point of the UI canvas.  I still need to:
-refactor SASS to use variables for repeated values (this is in progress on my end)
-eliminate grid layout since it will probably not be useful for showing and hiding topbar
-probably eliminate the entire grid layout component and just bake all of the into some divs and flex boxes in the render in app.tsx


cypress/integration/workspace.test.ts
```diff
@@ -5,7 +5,9 @@ context("Test the overall app", () => {
 
   describe("Desktop functionalities", () => {
     it("renders with text", () => {
-      cy.get(".app").should("have.text", "Hello World");
+      //cy.get(".app").should("have.text", "Hello World");
```
ekosmin: I'd delete this old code rather than comment it out

src/components/investigate-dialog.sass
```diff
@@ -0,0 +1,99 @@
...
+  border: 3px solid $component-bgcolor
+  border-radius: $dialog-border-radius
+  opacity: 1
+  -webkit-animation-name: fadein /* Safari 4.0 - 8.0 */
+  -webkit-animation-duration: .5s /* Safari 4.0 - 8.0 */
+  animation-name: fadein
+  animation-duration: .5s
+
+/* Safari 4.0 - 8.0 */
+@-webkit-keyframes fadein
```
ekosmin: `autoprefixer` should make these prefixes unnecessary, though I haven't tested it myself.

src/components/investigate-dialog.sass
```diff
@@ -0,0 +1,99 @@
...
+  from {opacity: 0;}
+  to {opacity: 1;}
+
+/* Standard syntax */
+@keyframes fadein
```
ekosmin: Animations look nice! Would be nice to have a standard for when things fade-in/fade-out throughout the app

src/components/investigate-dialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+interface IState {}
+
+@inject("stores")
+@observer
+export class InvestigateDialogComponent extends BaseComponent<IProps, IState> {
+
+  public render() {
+    return (
+      <div className="investigate-dialog">
+        <div className="investigate-dialog-header">
```
ekosmin: Normally I leave out the `investigate-dialog` prefixes here, and scope all my CSS rules so they only apply to the component they're styling. How do you feel about that idea? If we want to have headers like this, that would be an argument for using CSS Modules.

src/components/investigate-dialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+          </div>
+        </div>
+        <div className="investigate-dialog-footer"/>
+      </div>
+    );
+  }
+
+  private handleClickClose = () => {
+    const {ui} = this.stores;
+    { ui.setShowInvestigationModalSelect(!ui.showInvestigationModalSelect); }
```
ekosmin: Braces here and below seem unnecessary.

src/components/investigate-dialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  private handleClickInvestigatePopulation = () => {
+    const {ui} = this.stores;
+    { ui.setShowInvestigationPanel(true); }
```
ekosmin: One next step might be to update the model to support multiple investigation panels with different colors e.g. a cell zoom panel. You'll probably want to keep track of some kind of `enum` string here instead in that case.

src/components/main-content.tsx
```diff
@@ -0,0 +1,37 @@
...
+    );
+  }
+
+  private renderDialog() {
+    const {ui} = this.stores;
+    const {showInvestigationModalSelect} = this.stores.ui;
+    const {showInvestigationPanel} = this.stores.ui;
+    // {return ui.showInvestigationModalSelect ? "true" : "false"}
+    if (showInvestigationModalSelect) {
+      return <InvestigateDialogComponent/>;
```
ekosmin: Do we really want the modal to hide the active investigation? I'd think it would just darken it like in the mockup.

src/components/main-content.tsx
```diff
@@ -0,0 +1,37 @@
...
+    } else if (showInvestigationPanel) {
+      return <InvestigationPanelComponent/>;
+    } else {
+      return <div/>;
```
ekosmin: I'm not sure you need to return an empty div here

src/components/main-content.tsx
```diff
@@ -0,0 +1,37 @@
...
+        {this.renderDialog()}
+      </div>
+    );
+  }
+
+  private renderDialog() {
+    const {ui} = this.stores;
+    const {showInvestigationModalSelect} = this.stores.ui;
+    const {showInvestigationPanel} = this.stores.ui;
+    // {return ui.showInvestigationModalSelect ? "true" : "false"}
```
ekosmin: Dead code

src/models/ui.ts
```diff
@@ -2,12 +2,20 @@ import { types } from "mobx-state-tree";
...
+    sampleText: "Hello World",
+    showInvestigationModalSelect: false,
+    showInvestigationPanel: false
   })
   .actions((self) => {
     return {
       setSampleText(text: string) {
         self.sampleText = text;
+      },
+      setShowInvestigationModalSelect(val: boolean) {
```
ekosmin: They'll be trivial, but would you mind adding some tests to `ui.test.ts` for these new functions?

If you don't mind, it would be helpful to get rid of uses of `sampleText` too, now that we have actual functional examples

cypress/integration/workspace.test.ts
mklewandowski: Sounds good.  Will delete.

src/components/investigate-dialog.sass
mklewandowski: Just tested.  From the generated CSS, it looks like it works!

src/components/investigate-dialog.sass
mklewandowski: Agree.  We need to standardize times, etc.  Could make a var in `vars.sass`.  I think there is a story about animations so maybe best to attack it there?

src/components/main-content.tsx
mklewandowski: Will remove.

src/components/main-content.tsx
mklewandowski: We don't.  I originally had it showing text in the div for testing, but now it's just adding an unnecessary div.  Will remove.

src/components/investigate-dialog.tsx
mklewandowski: I must not be understanding proper syntax.  It throws and error without the braces.

src/components/investigate-dialog.tsx
mklewandowski: Agree.  This might be another good place for a story.  Break the investigation-panel apart so it can support displaying different types of content.

src/models/ui.ts
mklewandowski: Yup.  No problem.

src/components/main-content.tsx
mklewandowski: This entire chunk needs to be reworked to handle multiple investigation types and we need to add the semi-opaque curtain behind the dialog.  This might be best done in another story (possibly the same as the above comment).

src/components/investigate-dialog.tsx
ekosmin: In case that wasn't clear, I meant something like this instead:

```
<div className="investigate-dialog">
  <div className="header">
    <div className="title">
...
```

And then in the SASS:

```
.investigate-dialog
  foo: bar
  .header
    baz: qux
    .title
...
```

mklewandowski: Or perhaps I'm not following.  Could you clarify?

ekosmin: Ah sorry no you've got it right! Didn't see your message before I finished writing an example. I don't actually have a strong opinion... @sfentress or @dstrawberrygirl do you?

mklewandowski: We can try something like that.

src/components/investigate-dialog.tsx
ekosmin: Interesting, it compiles on my end - I just pushed up a commit and it seems like Travis handled it fine too. 

Braces are used in React to run code in a JSX object. So I could say something like this:

```
render() {
  return (
    const className = "name"
    <div className={className}>
      {"Welcome " + props.user ? urlParams.user : "User"}
      {props.showThing && <div>thing</div>}
    </div>
  )
}

src/components/investigate-dialog.tsx
mklewandowski: Yeah, that works fine.  l think it makes it a little harder to search the DOM or figure out where you are when you're inspecting elements (since class names might repeat or be too generic to tell much about an element), but I don't have any major issues with this.  I can easily refactor the SASS with this style.

mklewandowski: Anyway, I'm fine with either.  Both have advantages/disadvantages.  I can wait for others to chime in.

src/components/investigate-dialog.tsx
mklewandowski: Are we looking at different lines?  I'll look at your commit.

mklewandowski: Bleh.  I thought you meant here:
`const {ui} = this.stores;`

mklewandowski: Yup.  Works fine.

mklewandowski: And those braces, I assume, are needed as part of the MST syntax?

kswenson: Basic TS/JS destructuring -- not MST.

src/components/investigate-dialog.tsx
mklewandowski: Also, I kind of like the tighter scoping that you presented.  Seems fine (and probably more modern for complex styling).

src/components/investigate-dialog.tsx
ekosmin: Ah, yes! Sorry I didn't specify. And right, as @kswenson said:

`const {ui} = this.stores;`

is equivalent to:

`const ui = this.stores.ui;`

ekosmin: Ah yes, sorry my comment wasn't more specific!

Right, what Kirk said - basically this:

`const {ui} = this.stores;`

is equivalent to this:

`const ui = this.stores.ui;`

src/components/investigate-dialog.tsx
ekosmin: ^ I think your points about searching the DOM are good, it's definitely more difficult to select the element you want. Ultimately I think I am happier with the updated styles, since the best way to implement unique selectors would probably still be CSS Modules.

dstrawberrygirl: I think I'd rather go with the simplified styling idea. A header should have some common properties throughout the app for consistency. If we keep component sass definitions next to the component, as we did with CLUE, then it's clear that the styles we use for a particular component will be used and will override a common base. It's not hard to look up a tree and find the parent that helps pinpoint a particular style definition.  

Some elements will benefit from including a more verbose name - those that are more unique, those that are much less likely to share styling elsewhere. Those cases are fine to be more verbose.

mklewandowski: Agreed.  I think nesting the classes in Sass is what we settled on in the last update.  Should be less to type and allows generic elements to simply nest within a more uniquely named parent.

PR #3, opened 2018-11-08
URL Parameters

ekosmin:
Here's a first pass at encoded JSON URL parameters. They're parsed and stored on load and can be imported anywhere. 

@pjanik and @scytacki had discussed using `react-jsonschema-form` to generate authoring pages that could validate inputs and return a JSON object. I gave it a try and it was pretty easy to use - only difficulty is that TypeScript complains unless the form schema is defined in-line because of some confusion over string literals. I also had some trouble keeping the form's data in React state without causing render loops, but fortunately I don't need to do that for now.

### DIFF
--- a/cypress/integration/workspace.test.ts
+++ b/cypress/integration/workspace.test.ts
@@ -5,8 +5,8 @@ context("Test the overall app", () => {
 
   describe("Desktop functionalities", () => {
     it("renders with text", () => {
-      cy.get('.app-container').should('exist');
-      cy.get('.nav-and-content-container').should('exist');
+      cy.get(".app-container").should("exist");
+      cy.get(".nav-and-content-container").should("exist");
     });
   });
 });
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -7,6 +7,7 @@ import "./app.sass";
 import { TopBarComponent } from "./top-bar";
 import { LeftNavPanelComponent } from "./left-nav-panel";
 import { MainContentComponent } from "./main-content";
+import { urlParams } from "../utilities/url-params";
 
 interface IProps extends IBaseProps {}
 interface IState {}
@@ -16,10 +17,9 @@ interface IState {}
 export class AppComponent extends BaseComponent<IProps, IState> {
 
   public render() {
-    const {ui} = this.stores;
     return (
       <div className="app-container">
-        <TopBarComponent />
+        {urlParams.topBar && <TopBarComponent />}
         <div className="nav-and-content-container">
           <LeftNavPanelComponent />
           <MainContentComponent />
--- a/src/components/authoring.tsx
+++ b/src/components/authoring.tsx
@@ -1,0 +1,35 @@
+import * as React from "react";
+import { BaseComponent, IBaseProps } from "./base";
+
+import "./app.sass";
+import Form, { ISubmitEvent } from "react-jsonschema-form";
+import { AppMode } from "../models/stores";
+import { QueryParams } from "../utilities/url-params";
+
+interface IProps extends IBaseProps {}
+interface IState {}
+
+export class AuthoringComponent extends BaseComponent<IProps, IState> {
+  public render() {
+    const onSubmit = (e: ISubmitEvent<QueryParams>) => {
+      const encodedParams = encodeURIComponent(JSON.stringify(e.formData));
+      window.open(`${location.origin}${location.pathname}?${encodedParams}`, "connected-bio-spaces");
+    };
+    return (
+      <div className="authoring">
+        <Form
+          schema={{
+            title: "Connected Bio Parameters",
+            type: "object",
+            properties: {
+              appMode: {type: "string", title: "App Mode", default: "dev", enum: [
+                "dev",
+                "authed"
+              ]}
+            }
+          }}
+          onSubmit={onSubmit} />
+      </div>
+    );
+  }
+}
--- a/src/components/authoring.tsx
+++ b/src/components/authoring.tsx
@@ -22,6 +22,7 @@ export class AuthoringComponent extends BaseComponent<IProps, IState> {
             title: "Connected Bio Parameters",
             type: "object",
             properties: {
+              topBar: {type: "boolean", title: "Show Top Bar?", default: true},
               appMode: {type: "string", title: "App Mode", default: "dev", enum: [
                 "dev",
                 "authed"
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -6,12 +6,21 @@ import { AppComponent } from "./components/app";
 import { createStores } from "./models/stores";
 
 import "./index.sass";
+import { urlParams } from "./utilities/url-params";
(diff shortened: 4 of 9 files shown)


src/index.tsx
```diff
@@ -6,12 +6,21 @@ import { AppComponent } from "./components/app";
 import { createStores } from "./models/stores";
 
 import "./index.sass";
+import { urlParams } from "./utilities/url-params";
+import { AuthoringComponent } from "./components/authoring";
 
-const stores = createStores({ });
+const stores = createStores({ appMode: urlParams.appMode });
```
sfentress: Are we actually planning on using `appMode`? What's the use-case? If we're not, should we simplify things by getting rid of it?

src/components/app.tsx
```diff
@@ -16,10 +17,9 @@ interface IState {}
 export class AppComponent extends BaseComponent<IProps, IState> {
 
   public render() {
-    const {ui} = this.stores;
     return (
       <div className="app-container">
-        <TopBarComponent />
+        {urlParams.topBar && <TopBarComponent />}
```
sfentress: I feel like there's a potential for a spaghetti of different components asking about urlParams in a just-in-time kind of way. Did you consider using the url params once to set up the initial stores, and then only referencing everything through the stores?

E.g. `showTopBar` seems like it's a natural property of the `ui` store. If we created the stores in index.tsx with `const stores = createStores({ urlParams });`, then every store could grab the properties it needs and initialize itself. (We could even leverage the nesting of the JSON to automatically pass the relevant props to the relevant stores, maybe.)

Then `AppComponent` would just check the store to chose whether to display `<TopBarComponent />`.

Another potential advantage is that `AuthoringComponent` could be displayed by `AppComponent` (instead of switching to it in index.tsx) and then it would also have easy access to all the stores, and it could pre-populate the authoring form with the current values from the store. So if authors had a long url and they just wanted to tweak one thing, they would simply add `&authoring` to the url and see the form already set up. This would also work if we ever went with LARA authoring, where the JSON is stored on the backend, and an author would expect to see the correct form for a model they've already authored.

src/components/authoring.tsx
```diff
@@ -0,0 +1,36 @@
...
+      window.open(`${location.origin}${location.pathname}?${encodedParams}`, "connected-bio-spaces");
+    };
+    return (
+      <div className="authoring">
+        <Form
+          schema={{
+            title: "Connected Bio Parameters",
+            type: "object",
+            properties: {
+              topBar: {type: "boolean", title: "Show Top Bar?", default: true},
```
mklewandowski: Is hiding the top bar part of a larger embedded mode?  If so, is it better to have a singular embedded mode or let the user mix and match features with the authoring tool?   Either way, this seems fine for now.

src/index.tsx
mklewandowski: I had the same question.  Is `appMode`  a holdover from CLUE?

cypress/integration/workspace.test.ts
```diff
@@ -5,8 +5,8 @@ context("Test the overall app", () => {
 
   describe("Desktop functionalities", () => {
     it("renders with text", () => {
-      cy.get('.app-container').should('exist');
-      cy.get('.nav-and-content-container').should('exist');
+      cy.get(".app-container").should("exist");
+      cy.get(".nav-and-content-container").should("exist");
```
mklewandowski: Thanks for fixing this!  At one point, I was tinkering with Cypress syntax to see if it accepted both single and double quotes - must have overlooked restoring it.  I'm actually not a fan of single quotes unless they are absolutely needed.

src/index.tsx
ekosmin: Ah sorry yes, I'll get rid of this

src/components/authoring.tsx
ekosmin: Oh good question...agreed I'll leave it for now, but if more features depend on this it would be worth changing the URL parameter

mklewandowski: Agreed.  An embedded mode might be nothing more than hiding the topbar.  I did briefly work on some changes in CODAP's embedded mode (for when it gets placed in LARA or other apps) recently, and it has a variety of minor, nuanced things that it does when embedded.  Regardless, probably a question for another day.

src/components/app.tsx
ekosmin: I hadn't considered this initially! I chose this path because it's how URL parameters work in CLUE. I liked your idea though and got most of the way through implementing it before I had some second thoughts.

I agree that `showTopBar` is information that could go in the UI model, but as far as I know, the value will never change during execution. I'm not sure how I feel about adding an extra piece of state to the model just so that we can pull values from a store instead of from the URL params.

As far as rendering right in `index.tsx` vs. in `AppComponent` - I'm not sure how things work in LARA, but at the moment it would still be easy to support adding `&authoring` to bring up a pre-loaded form, just by grabbing the current URL params. It's definitely true that this wouldn't work if we wanted to pull values out of state, but I can't think of when we'd want to do that off the top of my head.

Let me know what you think, happy to talk more next week! @kswenson, also curious if you have any CLUE-related thoughts.

kswenson: I think @sfentress is right that reading URL parameters from arbitrary points in the code sets a bad precedent. I think it would make more sense for `index.tsx` to read the URL parameters and distribute them appropriately. Whether that distribution should be through the stores is a separate question, however. In this instance it would make sense for it to be passed as a prop to `AppComponent`, in which case whether or not it's in stores becomes an implementation detail.

src/components/authoring.tsx
kswenson: This is another reason to hoist URL processing to index.tsx. The AppComponent render function shouldn't have to change if we decide to change the URL parameter to a mode. Having `showTopBar` be a prop puts that responsibility where it makes more sense.

src/components/app.tsx
sfentress: I could see `showTopBar` being either in a prop or the store, but in any case I feel like many of the props will end up being things that will be in stores -- e.g. the initial mice in the backpack (users can might more), the initial tool being shown (users can change it), etc. Most of what an author will be doing will be setting state, I think, so I feel it makes sense to store it as state.

ekosmin: @kswenson @sfentress Good points all around. I just added a new commit that addresses concerns about references to `url-params`. Now, all URL processing is done in `index.tsx`. I passed `showTopBar` as a prop to the `AppComponent`, but would definitely pass params to store-creation in the future for things like `initialMice`.

One question I considered was where to put default values. I opted for putting them in `url-params` rather than adding default values at the point of use, which is more or less what we do in CLUE.

There are a lot of small decisions to make here - let's hop on a call if there are any more changes you'd like to see/consider.

ekosmin: Oh also - in a perfect world, I'd add an Enzyme test to make sure the App renders correctly with the new prop, but I want to get this merged sooner so that others can use it.

PR #4, opened 2018-11-08
add legend at bottom of left nav panel

mklewandowski:
This was one of Kiley's requests to knock out based on Michael's mock-ups.  Changes add a new legend React component that is placed at the bottom of the left nav panel.  This legend is a color key which indicates which colors are used for male, female, heterozygote organisms.

### DIFF
--- a/src/components/left-nav-panel.sass
+++ b/src/components/left-nav-panel.sass
@@ -46,6 +46,7 @@
       flex-grow: 1
       overflow-y: auto
       background-color: $left-nav-collect-bgcolor
+      margin-bottom: 0px
 
   .button
     text-align: center
@@ -72,6 +73,17 @@
   .button:hover
     border-color: $button-border-hover-color
 
+  .footer
+    flex-grow: 0
+    flex-shrink: 0
+    flex-basis: $left-nav-footer-height
+    display: flex
+    justify-content: center
+    align-items: center
+    margin-bottom: $component-margin
+    color: $left-nav-footer-color
+    background-color: $left-nav-footer-bgcolor
+
 
 /* SCROLLBAR */
 /* width */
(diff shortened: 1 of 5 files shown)


src/components/left-nav-panel.sass
```diff
@@ -72,6 +73,17 @@
   .button:hover
     border-color: $button-border-hover-color
 
+  .footer
+    flex-grow: 0
+    flex-shrink: 0
+    flex-basis: $left-nav-footer-height
```
ekosmin: This looks good - I've just used `height` in the past but it seems like this is more specific. Is it necessary to set `flex-grow` and `shrink` here?

mklewandowski: It seemed like the more "flexy" way to specify a constant piece within a set of flexible containers.  So we define the start dimension with `flex-basis` and then set `flex-grow` and `flex-shrink` to 0 to tell it not to change size.  It's verbose and basically the same as `height`, but I like using the flex properties to specify dimensions if we're using flex boxes.

ekosmin: Oh I just meant, the defaults on `flex-grow` and `flex-shrink` are already 0, right?

mklewandowski: Ah, good point!

mklewandowski: Looks like flex-shrink is 1 by default:
https://developer.mozilla.org/en-US/docs/Web/CSS/flex-shrink
Flex-grow is 0:
https://developer.mozilla.org/en-US/docs/Web/CSS/flex-grow

mklewandowski: We also need to worry about inherited values since this layout does use a number of flex boxes.

PR #5, opened 2018-11-08
161766223 two up display

mklewandowski:
Add a PopulationSpace component that displays 2 side-by-side panels which will eventually house the population sim on the right and a graph/data view on the left.  Remove the now obsolete InvestigationPanel component.  Display the PopulationSpace component when the user clicks on the population button in the investigation dialog.  Update the UI model to contain population graph state (this will need to be reworked to have greater complexity.  We could use an enum that reflects the current space that is visible.  This is probably best done in its own story and branch).


src/components/population-space.tsx
```diff
@@ -0,0 +1,60 @@
...
+  public render() {
+    return (
+      <div className="population-space">
+        {this.RenderLeftPanel()}
+        {this.RenderRightPanel()}
+      </div>
+    );
+  }
+
+  private RenderLeftPanel() {
```
ekosmin: Should probably be lowercased

src/components/population-space.tsx
```diff
@@ -0,0 +1,60 @@
...
+        <div className="header">
+          <div className="title">Investigate: Population</div>
+        </div>
+        <div className="content">content here</div>
+      </div>
+    );
+  }
+
+  private RenderRightPanel() {
+    const {ui} = this.stores;
```
ekosmin: Agreed about eventually refactoring state - we may want to create a separate UI store for each space

src/components/population-space.tsx
mklewandowski: Yes!  I got too caught up with the uppercase on the exported React component name.  Will change.

src/components/population-space.tsx
mklewandowski: Agree.  We'll need a story for that.

sfentress:
:+1: Looks good.

PR #6, opened 2018-11-09
Added data-test attributes for testing

kevinsantos-cc:


mklewandowski:
Looks like travis build failed here:
ERROR: src/components/investigate-dialog.tsx[22, 1]: Exceeds maximum line length of 120

src/components/left-nav-panel.tsx
```diff
@@ -16,21 +16,21 @@ export class LeftNavPanelComponent extends BaseComponent<IProps, IState> {
     return (
       <div className="left-nav-panel">
```
mklewandowski: We'll also want to add the legend from this PR:
https://github.com/concord-consortium/connected-bio-spaces/pull/4

src/components/investigate-dialog.tsx
```diff
@@ -13,32 +13,32 @@ export class InvestigateDialogComponent extends BaseComponent<IProps, IState> {
...
+      <div className="investigate-dialog" data-test="investigate-dialogue">
         <div className="header">
-          <div className="title">Investigate</div>
+          <div className="title" data-test="investigate-dialog-title">Investigate</div>
           <div className="close" onClick={this.handleClickClose}>x</div>
         </div>
         <div className="content">
-          <div className="message">Select a [space] to begin your investigation</div>
-          <div className="flex-container">
+          <div className="message" data-test="investigate-dialogue-message">Select a [space] to begin your investigation</div>
```
mklewandowski: This is the line that is causing the Travis build to fail.

src/components/investigation-panel.tsx
```diff
@@ -13,14 +13,14 @@ export class InvestigationPanelComponent extends BaseComponent<IProps, IState> {
 
   public render() {
     return (
-      <div className="investigation-panel">
```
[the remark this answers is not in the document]
mklewandowski: Anyway, as a result I removed this module entirely.  So we will need to add more attributes to the new population-model module.

cypress/integration/elements/InvestigateDialogue.js
```diff
@@ -0,0 +1,29 @@
+class InvestigateDialogue {
```
mklewandowski: For consistency, I would use "dialog" instead of "dialogue"

cypress/integration/elements/LeftNav.js
```diff
@@ -0,0 +1,15 @@
+class LeftNav {
+
+  getBackpack() {
```
mklewandowski: This is a good reminder of the term "backpack".  Do you think of the set of collected mice in the collect space on the left panel to be the "backpack?"  Just wondering if I should refactor any components to use the "backpack" term so others can more easily find the proper code.

cypress/integration/elements/workspace.js
```diff
@@ -0,0 +1,12 @@
+import LeftNav from './elements/LeftNav';
+import ViewBox from './elements/InvestigationDialogue';
```
mklewandowski: Same as above - I'd use "dialog."

src/components/investigate-dialog.tsx
```diff
@@ -13,32 +13,32 @@ export class InvestigateDialogComponent extends BaseComponent<IProps, IState> {
 
   public render() {
     return (
-      <div className="investigate-dialog">
+      <div className="investigate-dialog" data-test="investigate-dialogue">
```
mklewandowski: One more place where we should use "dialog" (there are a few more, but I'll stop littering this with redundant comments).

ekosmin:
Thanks for taking a look @mklewandowski! @kevinsantos-cc, I agree with what Matt had to say - especially about fixing up the >120 char line.

Also, my latest understanding with @eireland was that we would keep QA tests totally separate from dev tests, so I'm not sure if the additional Cypress tests belong here.

Once these pieces are cleaned up, I think it's fine to get this merged in. New fields can be labelled as needed for testing.

kevinsantos-cc:
I have made the changes requested and made the commits. First time doing PR's so let me know what else you might need from me!

PR #7, opened 2018-11-13
Add populations model space

sfentress:
This adds the basic populations model into the two-up display.

It first makes the 2-up display generic, by allowing it to take react components as properties.

It then sets up the basic populations model and view, and adds a simple external toolbar instead of the default populations.js toolbar. This will need to be styled in a later commit.

Several parameters are set by default on the model, like how the initial mouse population is defined. This will eventually get pulled from the authoring.


sfentress:
@mklewandowski I refactored the `TwoUpDisplayComponent` in this commit here: https://github.com/concord-consortium/connected-bio-spaces/pull/7/commits/6bd254

I rebased it onto the start of the PR to try and make it simpler.

I was not able to get TypeScript to be happy with child props. I think it would have looked nicer if `populations-space.tsx` could define its view as

```
<TwoUpDisplayComponent
  leftTitle="Investigate: Population"
  rightTitle={graphTitle}
  rightIcon={iconId}
  onClickRightIcon={this.togglePopulationsGraph}
  >
    <leftPanel>
      <PopulationsComponent />
    </leftPanel>
    <rightPanel>
      {graphPanel}
    </rightPanel>
</TwoUpDisplayComponent>
```
but I couldn't get TypeScript to be happy with it. Probably it doesn't matter.

In any case, let me know what you think.

mklewandowski:
Looks like Travis had some issues with the console logs and setting fillstyle on an invalid object, but it looks like you already checked in some changes.  I just pulled down the branch so I can test and take a look at code.

sfentress:
Yup, I've pushed the low-hanging fixes, but the failing test will take a little bit of thought.

src/components/main-content.tsx
```diff
@@ -22,13 +27,16 @@ export class MainContentComponent extends BaseComponent<IProps, IState> {
   }
 
   private renderDialog() {
```
mklewandowski: Probably time to change the name of this function.  renderSpace?  renderInvestigationContent?

src/components/main-content.tsx
```diff
@@ -22,13 +27,16 @@ export class MainContentComponent extends BaseComponent<IProps, IState> {
...
     const {showInvestigationPanel} = this.stores.ui;
+
+    // stawman code
+    const SpaceComponent = SpaceComponents[investigationPanelSpace];
+
     if (showInvestigationModalSelect) {
       return <InvestigateDialogComponent/>;
-    } else if (showInvestigationPanel) {
-      return <TwoUpDisplayComponent/>;
+    } else if (showInvestigationPanel && SpaceComponent) {
```
mklewandowski: If one of the values of type `Spaces` in the ui store is `none` then eventually we can probably whack `showInvestigationPanel` and instead just check if the current space is none (unless you intend none to have another meaning).

kswenson:
@sfentress The changes to `package-lock.json` seem excessive given the changes to `package.json`. What version of npm are you running?

src/components/two-up-display.tsx
```diff
@@ -24,37 +31,38 @@ export class TwoUpDisplayComponent extends BaseComponent<IProps, IState> {
...
       <div className="two-up-panel right-abutment">
         <div className="header">
-          <div className="button-holder" onClick={this.handleClickMode}>
-            <svg className="button">
-              <use xlinkHref={iconId} />
-            </svg>
-          </div>
-          <div className="title">{titleText}</div>
+          {
+            this.props.rightIcon ?
```
mklewandowski: Good idea to not assume we will have an icon that swaps panel state.

src/models/ui.ts
```diff
@@ -1,10 +1,13 @@
 import { types } from "mobx-state-tree";
 
+export type Spaces = "none" | "populations";
```
mklewandowski: @ekosmin and I discussed using an enumerated type for this (as is done in CLUE).  Might make the intellisense work better?  @ekosmin, do you have thoughts on this?

src/components/main-content.tsx
sfentress: `renderSpace` and `renderInvestigationContent` seem confusing when we *are* rendering the dialog. Maybe just `renderMainContent`?

src/components/main-content.tsx
mklewandowski: Right.  We don't have clarity on this.  What is displayed when we first bring up the app?  An empty panel?  An intro panel?  One of the spaces?  I liked including `none` since it more resembles a C family enum where `none` represents "no spaces shown", but if that is never a valid option.... Agree on close button too.  I suppose we need to ask the right person if we can ever have a state where no spaces are shown.

src/components/main-content.tsx
mklewandowski: I like it!  renderMainContent it is!

src/models/ui.ts
mklewandowski: I personally like what you did.  It's short and simple.  Just wanted to check if there is any reason to have CLUE and Cbio do the same thing in cases where we're "basically" trying to create an enumerated type.

sfentress:
@kswenson I've rebuilt package-lock using npm 6.4.1. This created so many more changes that GitHub refuse to show the diff. If we all settle on 6.4.1 though, hopefully this noise will be kept to a minimum.

kswenson:
Looking at the git history it appears that package-lock.json has been ping-ponging between npm 5/6 formats. @mklewandowski It looks like you should update to npm 6.4.1 as well.

PR #8, opened 2018-11-14
Chart foundations

dstrawberrygirl:
Add initial implementation of ChartJS into repository and demonstrate pulling data from the populations store to render alongside the populations space. Currently there are three types of chart to choose from - Line, Bar or Horizontal Bar, though adding additional types supported by ChartJS would likely not be hard.


package.json
```diff
@@ -104,6 +104,8 @@
   },
   "dependencies": {
     "@types/react-jsonschema-form": "^1.0.11",
+    "@types/chart.js": "^2.7.40",
```
kswenson: May not have any real consequences, but it's customary to put `@types/*` in `devDependencies` since they're only used at compile-time.

dstrawberrygirl: Good catch, thanks!

src/models/spaces/charts/chart-data.ts
```diff
@@ -0,0 +1,53 @@
...
+      self.data.forEach((d) => {
+        maxA1Values.push(d.maxA1 || 100);
+        maxA2Values.push(d.maxA2 || 100);
+        minA1Values.push(d.minA1 || 0);
+        minA2Values.push(d.minA2 || 0);
+      });
+
+      return {
+        maxA1: Math.min(...maxA1Values),
+        maxA2: Math.min(...maxA2Values),
```
sfentress: The two lines above should be `Math.max`, right? (I think the tests didn't catch it because there was only one data set.)

src/models/spaces/charts/chart-data-set.ts
```diff
@@ -0,0 +1,151 @@
...
+    get dataA2() {
+      if (self.data.length > 0 && self.data[0].a2) {
+        return self.data.map(p => p.a2);
+      } else {
+        return self.data.map(p => p.label);
+      }
+    },
+    // Determine minimum and maximum values on each axis
+    get maxA1(): number | undefined {
+      return Math.max(...self.data.map(p => p.a1));
```
sfentress: Note that if there is no data, `Math.max(...[])` is `-Infinity`, so the graph won't render. (Same with the other calculations below.) Maybe graphs should be required to have a default for when there is no data?

mklewandowski:
I'm loving the animations on the graph!  I noticed that when we resize the browser window, the graph and its container view will grow but never shrink.  This means that the two panels in the two-up display do not always consume 50% of the parent view width.  Is this intended?  Or do we want the two panels in the two-up display to always be 50% of the container view (and hence allow the graph to be scaled down as well as up)?

src/models/spaces/charts/chart-data-set.ts
```diff
@@ -0,0 +1,144 @@
...
+  name: string;
+  hex: string;
+}
+
+interface XYPoint {
+  x: number;
+  y: number;
+}
+// As defined in vars.sass, no easy way to get these variables in here
+// so reluctantly duplicating the definitions.
```
mklewandowski: I was hoping we could pull these out of the sass file.  Oh well.  Quick note, I started yanking a few generic colors from vars.sass and filling in spec colors defined here:
https://fh8yk2.axshare.com/#g=1&p=4_connectedbio_-_specs__v1_3_
If colors are duplicated for the chart in code, then we need to remember to pull some UI spec colors out for use in the graph (this can probably wait until Michael has a chance to review and offer color input).

sfentress: It looks like it should be possible to export the sass variables using the :export directive: https://til.hashrocket.com/posts/sxbrscjuqu-share-scss-variables-with-javascript

You can export variables from any style sheet: https://www.bluematador.com/blog/how-to-share-variables-between-js-and-sass

src/models/spaces/charts/chart-data.ts
```diff
@@ -32,6 +32,9 @@ export const ChartDataModel = types
         minA1: Math.min(...minA1Values),
         minA2: Math.min(...minA2Values),
       };
+    },
+    get nextDataSeriesColor() {
```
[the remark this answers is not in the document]
dstrawberrygirl: Have now added more colors, so we get up to 10 different dataset colors before we run out - unsure on the best course of action here on additional sets... more options or do we loop back to the beginning?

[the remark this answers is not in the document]
sfentress: I expect that the "standard" colors will end up being a nice-to-have, but ultimately unused, as for both populations and the cell model there has been a desire so far for colors to match the things in the model. So I think this isn't a problem.

PR #9, opened 2018-11-15
161596944 collect buttons

mklewandowski:
Adds a new CollectButton React component that can be used to display the various states of the collection buttons in the left nav panel. 
- There *might* be some final UI spec changes coming from Michael.  If so, I'll work those in as they come.  However, I want folks to start taking a look at this sooner rather than later (especially so we can think about storing the button state).
- At present, I'm pre-populating the left nav panel with 6 pre-defined CollectButton components.  I build an array of objects that contain the button specs and then build the buttons accordingly based on the array content.  Obviously we need to remove this temporary code and create the actual collect button state (this probably merits a discussion - are we storing buttons states, collected organism states, something else?  There a few different ways to think about this).  However, for testing/demoing the layout, these pre-populated buttons display the various button states.
- button events are hooked up and firing alerts at present.  They'll need to be modified to do something useful.  X-close seems fairly straight forward - it will remove a collected organism  But does a click on the circular collect button itself do anything?  We might not know until Michael gets a little further with the spec.


mklewandowski:
@sfentress let me rough in a quick backpack model and add it to the stores.  I'll, for now, just replicate what I have here so we'll still need to tinker with the code, but it will get us closer to where we need to be,

mklewandowski:
@sfentress I roughed in a backpack model which also contains a collected mouse model.  These will both need refinement, but for now I'm populating these with test values and then rendering the Collect button React components based on what we find in the store.  This means it should look the same as before (test button states being pre-populated on the left panel), but the state is maintained in a store.

ekosmin:
Just got done with a synchronous round of refactoring with @mklewandowski which solidifies a bit more of the `Mouse` state.

@sfentress I agree with what you said about trying to keep more complicated, model-specific state out of the main model. I'm imagining something like a `SubstanceManipulations` store to keep track of changes to mice. That model will also need to be the one that ultimately gets the color of a mouse, since genotype alone cannot predict color.

sfentress:
This all looks good to me. :+1:

I know we said we should just go ahead and use mice directly, and not worry about the making it generic for the other units yet (even though we'll probably be implementing a new unit sooner than we think...), the one minor gesture towards trying to contain all the unit-specific stuff might be to put the mice assets under a `curriculum/mouse/` subfolder, like I did for the populations asses. It just seems a little wrong to have the mouse images right next to the stuff we know is going to be generic, like the connected-bio-logo.

Speaking of which, I put all the assets for the populations model in `/public`, and later realized that maybe `/assets` made more sense. Is there any practical difference between the two? I think maybe the whole `/public/curriculum` folder could be moved over to `/assets/curriculum`.

But in any case, this all looks good.

mklewandowski:
I like having the image assets all reside in the `assets` folder, and I think making suitable subfolders (like `assets/curriculum` as you suggested) is a good idea.  But I'm potentially biased by the Unity model where an `assets` folder is the primary location for images, icons, logos, etc.  In the end, I'm open to anything. I think we already have some asset redundancy that we need to sort out, and I'm assuming the mice assets are all going to revised at some point.  Let's get this in and then make a general decision about where assets will live.

sfentress:
Yes, definitely keep them in assets. And we can move them into a curriculum subfolder after the PR.

PR #10, opened 2018-11-19
Chart improvements

dstrawberrygirl:
This is a small PR to ensure we always retain all data from a data set, and improves handling of scaling on charts with some additional optional parameters to datasets:
fixedMin: can set to 0 to always start Y at 0
fixedMax: can set to a known quantity (i.e. 100) to set a constant scale
expandOnly: not compatible with fixedMax, which would override this value, this will ensure that scrolling horizontal line charts don't constantly rescale the upper bounds of the y axis

### DIFF
--- a/src/models/spaces/charts/chart-data-set.ts
+++ b/src/models/spaces/charts/chart-data-set.ts
@@ -68,61 +68,71 @@ export const ChartDataSetModel = types
     name: types.string,
     dataPoints: types.array(DataPoint),
     color: types.string,
-    maxPoints: types.number
+    maxPoints: types.maybe(types.number)
   })
+  .views(self => ({
+    get dataTail() {
+      if (self.maxPoints && self.maxPoints > 0 && self.dataPoints.length >= self.maxPoints) {
+        return self.dataPoints.slice(-self.maxPoints);
+      } else {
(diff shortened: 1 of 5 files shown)


src/models/spaces/charts/chart-data-set.ts
```diff
@@ -166,13 +194,19 @@ export const ChartDataSetModel = types
       self.dataPoints.splice(0, self.dataPoints.length);
     }
 
+    function setMaxDataPoints(maxPoints: number) {
+      self.maxPoints = maxPoints;
+    }
```
sfentress: With the maxPoints function above (and also eventually if you allowed setting fixedMin and fixedMax) there's no way to unset the value once set, right?

The behavior of the graph with this is unset (showing all data) may be a desirable one to switch to, but I don't think you can switch back to it once you switch away.

sfentress: Oh, actually I see that you could switch to auto-scaling the x-axis if maxPoints were zero. This is slightly hidden.

Maybe one option would be to have the type of maxPoints be `types.optional(types.number, 0)`, to show there is a default. Not sure if that is any more self-documenting.

PR #13, opened 2018-11-21
161784877 investigate panel

mklewandowski:
Add explore buttons (previously known as investigation buttons) to the upper left nav panel which allow a user to select which "space" is currently displayed.


src/models/ui.ts
```diff
@@ -1,23 +1,15 @@
 import { types } from "mobx-state-tree";
 
-export type Spaces = "populations";     // eventually a union of all the spaces
+export type Spaces = "populations" | "breeding" | "organism" | "dna" | "none";
```
ekosmin: Realizing this should probably just be `Space`

src/components/explore-button.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+@inject("stores")
+@observer
+export class ExploreButtonComponent extends BaseComponent<IProps, IState> {
+
+  public render() {
+    const {ui} = this.stores;
+    let classSuffix: string = ((ui.investigationPanelSpace === this.props.space) ? "active" : "clickable");
+    switch (this.props.space) {
+      case "populations":
```
ekosmin: I might prefer a map or just using `space + suffix` here. At the very least, I'd include a `default` case if you go for a switch statement

src/components/main-content.tsx
```diff
@@ -3,14 +3,16 @@ import * as React from "react";
 import { BaseComponent, IBaseProps } from "./base";
 
 import "./main-content.sass";
-import { InvestigateDialogComponent } from "./investigate-dialog";
 import { PopulationsSpaceComponent } from "./spaces/populations-space";
 
 interface IProps extends IBaseProps {}
 interface IState {}
 
 const SpaceComponents: any = {
```
ekosmin: This should probably be typed with something like:

```
type SpaceTypeToComponent = {
  [key in Space]: SpaceComponent | undefined
}
```

Which would presume we have some kind of union'd type for all SpaceComponents (just `PopulationsSpaceComponent` at the moment).

Also, I prefer to capitalize constants as SPACE_COMPONENTS, though we haven't been consistent on this.

@sfentress, do you have any thoughts on this?

src/models/ui.ts
mklewandowski: Good call.  `Space` is more correct for an enum-like type such as this.

src/components/explore-button.tsx
mklewandowski: Unifying the CSS and space names is probably the easiest (as long as we're om with the assumption that the CSS class will match the space name.  I am.  Will change.

src/models/ui.ts
mklewandowski: Although wouldn't it be better to handle it like the other enums?  And do something like:
'export const SpaceTypeEnum = types.enumeration("type", ["populations", "breeding", "organism", "dna", "none"]);'
'export type SpaceType = typeof SpaceTypeEnum.Type;'

src/components/main-content.tsx
mklewandowski: Ah, how does that work?  So we're specifying that the key has to be of type Space and the key's value has to be of type SpaceComponent (or undefined)?  How do we map the keys to the desired component?

mklewandowski: Or is the idea that we then make an object of type SpaceTypeToComponent and do the mapping there with  typescript enforcing the nice and neat key/value type rules for us?

src/models/ui.ts
```diff
@@ -1,19 +1,21 @@
...
 
+export const SpaceTypeEnum = types.enumeration("type", ["populations", "breeding", "organism", "dna", "none"]);
+export type SpaceType = typeof SpaceTypeEnum.Type;
 export type Space = "populations" | "breeding" | "organism" | "dna" | "none";
 
 export const UIModel = types
   .model("UI", {
     showPopulationGraph: false,
-    investigationPanelSpace: "none",
+    investigationPanelSpace: SpaceTypeEnum,
```
ekosmin: I think you might be able to do something like `"none" as SpaceTypeEnum` here if you wanted

src/models/ui.ts
ekosmin: Oh yes, this is definitely an improvement! Before this, the space type in the UI was just a string. After your change you ended up with 3 types, I think you want something like this:

```
export const Space = types.enumeration("type", ["populations", "breeding", "organism", "dna", "none"]);
export type SpaceType = typeof Space.Type;
```

src/components/main-content.tsx
ekosmin: Yes, the second one! So once we had the type we could do:

```
const spaceTypeToComponent: SpaceTypeToComponent = {
  populations: PopulationsSpaceComponent
  dna: PopulationsSpaceComponent
  ...
  none: undefined
}
```

Also, to make the type def above that I worked we would use something like:

```
type SpaceComponent = PopulationSpaceComponent; // Eventually union all space components
```

`[key in Space]` is called a mapped type, and has the added benefit of making sure that *every* element in space is mapped to a component. So if you add a new `Space`, you'll get a compiler error to remind you to expand the map. You can write `[key in Space]?` instead if you don't want all types mapped - this just means that you'll need to check for undefined outputs from the map.

Happy to discuss more tomorrow if you'd like!

src/models/ui.ts
mklewandowski: Oh, that's nice too!

src/components/main-content.tsx
mklewandowski: Anyway, I see the idea.  I'm getting one, probably simple, error.  I DMed you with more info.

kswenson: On the topic of formatting constants, it's obviously an area over which strong opinions have been voiced over the years. Personally, I find the `ALL_CAPS_CONSTANT` formatting approach makes it seem like the constants are yelling at and potentially overshadowing the surrounding code. I would lean to a more soft-spoken convention like `kCamelCaseConstant`, but I'm fine with either.

ekosmin: You know, especially now that most of our variables are constants anyway, all caps might be overkill. Is it possible that we don't need either?

Anyway, I'm fine sticking with `kCamelCase` for now. But please read all my code as yells to compensate.

src/models/ui.ts
mklewandowski: Oops!  Meant to rip that obsolete `type Space` out!  Thanks for catching that.  I prefer using the naming conventions that match the rest of the code (i.e., `SpaceTypeEnum` and `SpaceType` - although this is starting to feel very sci fi - and names are probably going to drift as we add more types).

src/components/main-content.tsx
mklewandowski: Also prefer the camelCase constant - the all caps is a little intrusive when constants start piling up (which in ES6 is everywhere).

src/models/ui.ts
ekosmin: d'oh, good call, looks like that is our convention after all! I forgot I changed it for `Genotype` since `GenotypeType` just sounds...weird.

src/components/main-content.tsx
kswenson: I would reserve the constant formatting for true constants. ES6 constants, e.g. `const { foo } = bar;` are `const` in the sense that they don't change during the life of a single function, but they're expected to have different values at different times in the running of the application.

src/models/ui.ts
mklewandowski: I ripped out the old `type Space`.  I kept the naming convention on the others, but if sticking `Type` on the end of the exported types gets too annoying/repetitive, we can change.

src/components/explore-button.tsx
mklewandowski: This is fixed.

PR #14, opened 2018-11-26
Cell model

ekosmin:
Here are the first steps of integration for the cell model - I squashed together a few commits to make it easier to review. The behavior is roughly the same as what was demoed this morning.

I'll take another look through now and add any thoughts/questions I had. Once this is merged we can start new features off master.


src/components/spaces/cell-zoom/OrganelleWrapper.tsx
```diff
@@ -341,28 +357,28 @@ export class OrganelleWrapper extends React.Component<OrganelleWrapperProps, Org
...
-  //       const visibleModes = this.organelleSelectorInfo[organelle].visibleModes;
-  //       return !visibleModes || visibleModes.indexOf(rootStore.mode) > -1;
-  //     });
-  //   return possibleTargets.find((t) => {
-  //     return evt.target._organelle.matches({selector: this.organelleSelectorInfo[t].selector});
-  //   });
-  // }
+  private getOrganelleFromMouseEvent(evt: any): OrganelleType | undefined {
+    const {cellZoom} = this.stores;
+    const possibleTargets: OrganelleType[] = (Object.keys(this.organelleSelectorInfo) as OrganelleType[])
```
ekosmin: This looks like another possible place to use the new enum pattern I suggested on Slack

src/models/spaces/cell-zoom/cell-zoom.ts
```diff
@@ -14,35 +16,80 @@ export const Organelle = types.enumeration("type", [
 ]);
 export type OrganelleType = typeof Organelle.Type;
 
+export const kSubstanceNames = [
```
ekosmin: Here is the enum code I was referencing on Slack

src/organelles/dots.yml
```diff
@@ -0,0 +1,27 @@
+---
```
sfentress: Should all the `.yml` files be cell-zoom space folder?

src/components/spaces/cell-zoom/OrganelleWrapper.tsx
```diff
@@ -0,0 +1,388 @@
...
+        percentLightness = 1;
+      }
+
+      // go from lightest to darkest in HSL space, which provides the best gradual transition
+
+      // lightest brown: rgb(244, 212, 141) : hsl(41°, 82%, 75%)
+      // darkest brown:  rgb(124, 81, 21)   : hsl(35°, 71%, 28%)
+
+      const light = [41, 82, 75];
+      const dark = [35, 71, 28];
```
dstrawberrygirl: I realize we're deep into migrated code at this point, but a general related question - are the colors for mice going to be globally defined? If so, what's the best color format for consistency and ease-of-use? Start at RGB and have helpers to provide conversion to HSL & hex values?

src/components/spaces/cell-zoom/OrganelleWrapper.tsx
```diff
@@ -94,8 +96,8 @@ export class OrganelleWrapper extends BaseComponent<OrganelleWrapperProps, Organ
 
     modelDef.container = {
       elId: this.props.elementName,
-      width: 500,
-      height: 312
+      width: MODEL_WIDTH,
+      height: MODEL_HEIGHT
```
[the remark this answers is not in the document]
ekosmin: I've never used `SizeMe` before but it always looks very useful...I just added a PT story to handle scaling and a note to follow the pattern in the populations model.

src/organelles/dots.yml
ekosmin: Good thinking, just moved everything

src/components/spaces/cell-zoom/OrganelleWrapper.tsx
ekosmin: Sounds like a good idea for mouse colors to get a global definition...could be nice to share colors across graphs and cells, for example.

src/components/two-up-display.tsx
```diff
@@ -36,6 +37,7 @@ export class TwoUpDisplayComponent extends BaseComponent<IProps, IState> {
         <div className="content" data-test="two-up-content">
           {this.props.leftPanel}
         </div>
+        <ZoomControl />
```
sfentress: @dstrawberrygirl I just realized that this adds zoom controls to all 2-up displays, including the population view. Depending on the window size, this can cover the populations control.

Would it make sense to add it to the cell-zoom space instead?

dstrawberrygirl: @sfentress Good call - I have a branch for basic zoom control implementation, so I'll make the change in there.

PR #15, opened 2018-11-27
Multiple Cell Views

ekosmin:
Per our discussion today, I'm starting on click-to-add mice to the cell view, starting with the ability to show multiple mice at a time. Let me know what you think of the updated state, which looks something like this:

The `CellZoom` model now contains a map of "cell mice" and an array of `CellZoomRow`s. Each row references a mouse from the map of mice - this way we can synchronize updates across views more easily. Most of the state that was previously tracked in the `CellZoom` model (e.g. hovered organelle, active assays) are now stored per-row instead.

I also quickly added cell color so you can see that the cell views are rendering different organisms.

### DIFF
--- a/src/components/spaces/cell-zoom/cell-mouse.ts
+++ b/src/components/spaces/cell-zoom/cell-mouse.ts
@@ -1,9 +1,12 @@
 import { types, Instance } from "mobx-state-tree";
 import { MouseColor } from "../../../models/mouse";
-import { SubstanceType, OrganelleType, kOrganelleInfo } from "../../../models/spaces/cell-zoom/cell-zoom";
+import { SubstanceType, kOrganelleInfo } from "../../../models/spaces/cell-zoom/cell-zoom";
+import { v4 as uuid } from "uuid";
+import { OrganelleType } from "../../../models/spaces/cell-zoom/cell-zoom-row";
 
 export const CellMouseModel = types
-  .model("Populations", {
+  .model("CellMouse", {
+    id: types.optional(types.identifier, () => uuid()),
     baseColor: MouseColor
   })
   .views(self => ({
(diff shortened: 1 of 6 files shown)


src/models/spaces/cell-zoom/cell-zoom-row.ts
```diff
@@ -0,0 +1,68 @@
+import { types, Instance } from "mobx-state-tree";
+import { ChartDataModelType, ChartDataModel } from "../charts/chart-data";
+import { DataPoint, ChartDataSetModel, ChartColors, DataPointType } from "../charts/chart-data-set";
+import { CellMouseModel } from "../../../components/spaces/cell-zoom/cell-mouse";
+import { kSubstanceNames } from "./cell-zoom";
+
+export const Organelle = types.enumeration("type", [
```
ekosmin: I moved a couple of enumerations into this file, even though I'd prefer to keep them all at the top level `CellZoomModel`. If I didn't move them, MST wasn't picking them up properly somehow, even when they were imported...perhaps one of you could help me figure out what's going on there?

sfentress: Been playing around with it and I can't work it out either. It seems like importing `Organelle` from another file should be exactly the same was when `cell-mouse.ts` imports `MouseColor`:

```
export const CellMouseModel = types
  .model("CellMouse", {
    id: types.optional(types.identifier, () => uuid()),
    baseColor: MouseColor
  })
```

where `MouseColor` is defined elsewhere as 

```
export const MouseColor = types.enumeration("type", ["brown", "white", "tan"]);
```

I'm just not seeing the difference between these approaches.

src/components/spaces/cell-zoom/cell-mouse.ts
```diff
@@ -1,16 +1,41 @@
 import { types, Instance } from "mobx-state-tree";
```
sfentress: How come the cell-mouse model is in src/components?

ekosmin: You're right it totally shouldn't be! Let me fix that in my new branch, just in case I can avoid some conflicts with this one

PR #16, opened 2018-11-27
Bar chart colors

dstrawberrygirl:
Bar charts changed to  display each data series as a different color, plus moved color helper functions to separate utility file and adjusted line rendering to use new helpers. I'm not sure at this time whether we should retain the ability to show datasets as one color on a bar chart - I haven't seen a design requirement for this as yet so I've left it out for now.

### DIFF
--- a/src/components/charts/bar-chart.tsx
+++ b/src/components/charts/bar-chart.tsx
@@ -55,15 +55,7 @@ const defaultOptions: ChartOptions = {
 const barDatasetDefaults: ChartData<any> = {
   label: "",
   fill: false,
-  pointBackgroundColor: "#fff",
-  pointBorderWidth: 1,
-  pointHoverRadius: 5,
-  pointHoverBorderWidth: 2,
-  pointRadius: 1,
-  pointHitRadius: 10,
   data: [0],
-  backgroundColor: ChartColors.map(c => hexToRGBValue(c.hex, 0.4)),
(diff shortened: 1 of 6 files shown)


ekosmin:
I think this looks like a nice refactor! 

I must be missing something though - I see that the bars are difference colors now, but I don't see the change that caused this to happen. Since the model is put together like this, I expected all the bars to stay blue:
```
chartDataSets.push(ChartDataSetModel.create({
        name: "Sample Dataset1",
        dataPoints: points,
        color: ChartColors[0].hex,
        maxPoints: 100
      }));
```

src/components/charts/bar-chart.tsx
```diff
@@ -59,18 +61,17 @@ const barDatasetDefaults: ChartData<any> = {
   pointHoverBorderWidth: 2,
   pointRadius: 1,
   pointHitRadius: 10,
-  data: [0]
+  data: [0],
+  backgroundColor: ChartColors.map(c => hexToRGBValue(c.hex, 0.4)),
```
sfentress: I'm not sure I quite understand how `backgroundColor` is an array of colors, but maybe it would be easiest to see this in use.

dstrawberrygirl:
Updated the branch in line with @ekosmin's comment to ensure that specifying a color has the effect of using that color for the bars. Color property is now optional, so the default if none is specified is to use multicolored bars. The same logic would work for a scatter plot, if we decided we needed one down the line.

PR #17, opened 2018-11-27
Add mice to backpack from Populations model

sfentress:
This implements click-to-select addition to the backpack.

http://connected-bio-spaces.concord.org/branch/add-mice-to-backpack/

In the populations model:

1. Add mice
2. Click Select
3. Model will pause if it is running. If it had previously been running, model will restart if user goes out of selection mode
4. Up to 6 mice can be selected. They will appear in the backpack and removed from the model. Mice can be removed from backpack to clear spaces

The selection button UI is placeholder.

This also improves scaling. Mostly at a code level, but this can also be seen because

1. If the window is small when the populations model is loaded, it will now start small (previously it was limited to a minimum of 450px wide)
2. Model resizing behavior is still not perfect. If model is running, it should resize up and down. If it is stopped, it will still only grow, not shrink. This seems like a low-priority bug, as users will generally not be resizing embedded model.
3. Model still cannot letterbox to the window height. I can't get a height out of `SizeMe`.This seems like a low-priority bug, as we will hopefully be getting fixed aspect ratio eventually.

https://www.pivotaltracker.com/story/show/161996047
https://www.pivotaltracker.com/story/show/161766270


src/components/collect-button.tsx
```diff
@@ -51,15 +51,9 @@ export class CollectButtonComponent extends BaseComponent<IProps, IState> {
     );
   }
   private handleClickButton = () => {
-   this.addTestMouseToBackpack();
-  }
-  private addTestMouseToBackpack = () => {
```
mklewandowski: Is it worth leaving the addtestMouseToBackpack function around as a potential tool for future use?  I'm ok either way, but it could be useful to have a function that adds a test mouse to the backpack.

sfentress: Sure, we can leave it. I'm not sure that the button view will be the best place to leave the orphaned function, but we we could leave it regardless.

sfentress:
Yes, we will need an asset re-org soon. It's hard because every PR is adding its own assets, so its hard to do so without treading on toes. But when that's quieted down, and/or when we put in the new art, we should try to unify.

As for the test, I could add a new test to test the max-backpack-slots. There wasn't actually a test of that before, right? The previous test was just checking that we can set the value in MST, but that was basically just testing that the MST `create` function works as advertised. We could now test the max slots, because there is now logic enforcing max slots.

sfentress:
@mklewandowski Added a test.

mklewandowski:
Right!  Looks good!  Thanks for adding these!

PR #18, opened 2018-11-28
Add authoring to enable/disable Investigation Spaces

mklewandowski:
Add authoring to enable/disable Investigation Spaces.  Spaces can be shown or hidden from the authoring page (use URL param "authoring") and a default displayed space can also be chosen.  By default, all spaces are enabled and no space is shown.


src/models/ui.ts
```diff
@@ -3,11 +3,31 @@ import { types } from "mobx-state-tree";
 export const SpaceTypeEnum = types.enumeration("type", ["populations", "breeding", "organism", "dna", "none"]);
 export type SpaceType = typeof SpaceTypeEnum.Type;
 
+export function createUIModel(authoring: any): UIModelType {
+  if (!authoring.showPopulationSpace && authoring.displayedSpace === "populations" ||
+      !authoring.showBreedingSpace && authoring.displayedSpace === "breeding" ||
+      !authoring.showOrganismSpace && authoring.displayedSpace === "organism" ||
+      !authoring.showDNASpace && authoring.displayedSpace === "dna") {
+    authoring.displayedSpace = "none";
```
sfentress: Really minor, but I find it confusing to modify the original authoring hash here, instead of using `let displayedSpace = authoring.displayedSpace; if.... displayedSpace = "none" `. But that's really not important, unless we ever let the author switch back and forth between model and authoring.

mklewandowski: I was on the fence with this too.  I wasn't crazy about modifying the original authoring param because it could potentially throw off debugging/monitoring/etc.  I'm going to tweak it by making a local variable that stores the displayedSpace.

PR #19, opened 2018-11-29
Backpack Selection

ekosmin:
Adds the ability to select mice from the backpack and put them into the cell-zoom view. I have a few questions about how the cell views maintain their state and around component reuse that I'd like to touch base on at some point. I'll leave more specific comments in the code.

@sfentress I'm hoping you can take the lead on this review; @mklewandowski I tagged you to take a look at the `collect-button` changes if you have a chance

### DIFF
--- a/src/components/collect-button.tsx
+++ b/src/components/collect-button.tsx
@@ -24,12 +24,13 @@ export class CollectButtonComponent extends BaseComponent<IProps, IState> {
     );
   }
   private renderFull(mouse: MouseType) {
-    const buttonClass = mouse.sex === "male" ? "collect-button male" : "collect-button female";
+    let buttonClass = mouse.sex === "male" ? "collect-button male" : "collect-button female";
+    buttonClass += this.isDeselected() ? " deselected" : "";
     const innerOutlineClass: string = mouse.isHeterozygote ? "inner-outline heterozygote" : "inner-outline";
     return (
       <div>
         <div className="collect-button-outline">
-          <div className={buttonClass}>
+          <div className={buttonClass} onClick={this.handleClickMouse}>
             <div className={innerOutlineClass}>
               <img src={mouse.baseImage} className="icon"/>
               <div className="label">{mouse.label}</div>
@@ -69,5 +70,20 @@ export class CollectButtonComponent extends BaseComponent<IProps, IState> {
   private handleClickRemove = () => {
     const {backpack} = this.stores;
     backpack.removeCollectedMouse(this.props.index);
+  }
+
+  private handleClickMouse = () => {
+    const { backpack } = this.stores;
+    const { index } = this.props;
+    if (backpack.activeSlot === index) {
+      backpack.deselectSlot();
+    } else {
+      backpack.selectSlot(index);
+    }
+  }
+  private isDeselected = () => {
+    const { backpack } = this.stores;
+    const { index } = this.props;
+    return backpack.activeSlot != null && backpack.activeSlot !== index;
   }
 }
--- a/src/components/collect-button.tsx
+++ b/src/components/collect-button.tsx
@@ -20,8 +20,10 @@ export class CollectButtonComponent extends BaseComponent<IProps, IState> {
   public render() {
     const { backpack } = this.stores;
     const { backpackMouse } = this.props;
+    let className = "collect-button-holder";
+    if (!backpackMouse) className += " uncollected";
     return (
-      <div className="collect-button-holder">
+      <div className={className}>
         {backpackMouse ? this.renderFull(backpackMouse) : this.renderEmpty()}
       </div>
     );
@@ -69,7 +71,7 @@ export class CollectButtonComponent extends BaseComponent<IProps, IState> {
     );
   }
   private handleClickButton = () => {
-   this.addTestMouseToBackpack();
+    // do nothing by default
   }
   private addTestMouseToBackpack = () => {
     const {backpack} = this.stores;
--- a/src/components/collect-button.tsx
+++ b/src/components/collect-button.tsx
@@ -3,10 +3,10 @@ import * as React from "react";
 import { BaseComponent, IBaseProps } from "./base";
 
 import "./collect-button.sass";
-import { Mouse, MouseType, UNCOLLECTED_IMAGE } from "../models/mouse";
+import { BackpackMouse, BackpackMouseType, UNCOLLECTED_IMAGE } from "../models/backpack-mouse";
 
 interface IProps extends IBaseProps {
-  backpackIndex?: number;
+  backpackMouse?: BackpackMouseType;
   clickMouse?: () => void;
   clickEmpty?: () => void;
   clickClose?: () => void;
@@ -19,16 +19,17 @@ export class CollectButtonComponent extends BaseComponent<IProps, IState> {
 
   public render() {
     const { backpack } = this.stores;
-    const { backpackIndex } = this.props;
-    const mouse = backpack.getMouseAtIndex(backpackIndex);
+    const { backpackMouse } = this.props;
     return (
       <div className="collect-button-holder">
-        {mouse ? this.renderFull(mouse) : this.renderEmpty()}
+        {backpackMouse ? this.renderFull(backpackMouse) : this.renderEmpty()}
       </div>
     );
   }
-  private renderFull(mouse: MouseType) {
-    const { backpackIndex, clickMouse, clickClose } = this.props;
+  private renderFull(mouse: BackpackMouseType) {
+    const { backpack } = this.stores;
+    const { backpackMouse, clickMouse, clickClose } = this.props;
+    const backpackMouseIndex = backpack.getMouseIndex(backpackMouse);
     const handleMouse = clickMouse ? clickMouse : this.handleClickMouse;
     const handleClose = clickClose ? clickClose : this.handleClickRemove;
     let buttonClass = mouse.sex === "male" ? "collect-button male" : "collect-button female";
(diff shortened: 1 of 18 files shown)


src/components/collect-button.tsx
```diff
@@ -6,8 +6,10 @@ import "./collect-button.sass";
 import { Mouse, MouseType, UNCOLLECTED_IMAGE } from "../models/mouse";
 
 interface IProps extends IBaseProps {
-  mouse?: MouseType;
-  index: number;
+  backpackIndex?: number;
```
ekosmin: I made a handful of changes to `collect-button` - first, I removed the `mouse` from the props, since the `backpackIndex` alone is enough to determine the mouse. The more substantial changes were to allow me to reuse the buttons in the cell zoom UI. I want the buttons to look exactly the same, but the behavior is quite different so I added click handler overrides.

In the end I'm not sure it was worth adding these complications (e.g. allowing for an undefined `backpackIndex`), but I wanted to avoid copy/pasting the entire file. Let me know if you have any suggestions.

src/models/spaces/cell-zoom/cell-zoom-row.ts
```diff
@@ -19,12 +19,17 @@ export type OrganelleType = typeof Organelle.Type;
 export const Mode = types.enumeration("type", ["add", "subtract", "assay", "normal"]);
 export type ModeType = typeof Mode.Type;
 
+export const ZoomLevel = types.enumeration("type", ["organism", "cell", "protein"]);
+export type ZoomLevelType = typeof ZoomLevel.Type;
```
ekosmin: I added a `ZoomLevel` here because I thought I'd be implementing the organism view...it's not in use yet but I figured I'd leave it for reference.

src/models/spaces/cell-zoom/cell-zoom.ts
```diff
@@ -82,8 +83,48 @@ export const kOrganelleInfo: OrganelleInfo = {
 
 export const CellZoomModel = types
   .model("CellZoom", {
-    organisms: types.map(CellMouseModel),
-    rows: types.array(CellZoomRowModel)
+    cellOrganisms: types.optional(types.array(types.maybe(CellMouseModel)),
```
ekosmin: This is the piece of state I'd like to discuss. As mice are selected for the cell view, I convert the "backpack mouse" representation into a "cell mouse" representation and store it in this array by "backpack index". If the same mouse is added to a second view, it uses the stored "cell mouse" rather than create a new one. The intention was to simplify substance additions, so that if the substances/color change in one cell view they're reflected in the other. "Cell mice" are only removed from the array when no more views show that backpack mouse.

There were some complications here - I originally wanted to store the "cell mice" in a map, but MST *only* lets you store objects with `id`s by `id`. Storing in an array is fine, except you can't set a value greater than the array length so I needed to initialize it with 6 `undefined` values.

src/components/collect-button.tsx
sfentress: I think this makes sense. I think a clearer solution *might* have been to make an `OrganismCircle` component which handles the image and has generic click handlers, and then `BackpackItem` component would be a very thin wrapper on top of that, and the `CellZoom` would bring it in directly. But I think that this is fine.

src/components/collect-button.tsx
```diff
@@ -55,13 +65,36 @@ export class CollectButtonComponent extends BaseComponent<IProps, IState> {
...
+    const {backpackIndex} = this.props;
+    if (backpackIndex != null) {
+      backpack.removeCollectedMouse(backpackIndex);
+    }
+  }
+
+  private handleClickMouse = () => {
+    const { backpack } = this.stores;
+    const { backpackIndex } = this.props;
+    if (backpack.activeSlot === backpackIndex) {
```
sfentress: This could also be handled in the model logic. The view could just call `backpack.toggleSelection(backpackIndex)`.

src/components/spaces/cell-zoom/cell-zoom.tsx
```diff
@@ -0,0 +1,57 @@
...
+  private clickMouse = () => {
+    // Clicks only work on empty slots
+  }
+
+  private clickEmpty = () => {
+    const { backpack, cellZoom } = this.stores;
+    const { rowIndex } = this.props;
+    const { activeSlot } = backpack;
+    if (activeSlot != null) {
+      cellZoom.setRowBackpackIndex(rowIndex, activeSlot, backpack);
```
sfentress: :+1:

src/models/spaces/cell-zoom/cell-zoom-row.ts
sfentress: I think it's nice to have the enumeration. I don't know if it will cause conflicts when merging in with @dstrawberrygirl 's changes, but I think it's good to have the explicit words. If we eventually have multiple targets to zoom into, the model might even be able to set the options of this list.

src/components/collect-button.tsx
```diff
@@ -16,33 +18,41 @@ interface IState {}
 export class CollectButtonComponent extends BaseComponent<IProps, IState> {
 
   public render() {
-    const { mouse } = this.props;
+    const { backpack } = this.stores;
+    const { backpackIndex } = this.props;
```
sfentress: Oh, one problem with just storing the index and not the mouse itself, and using this same view in your zoom model: If the user is looking at a mouse cell, and then deletes the mouse in the backpack, the next mouse down is now at that index, and that's what the circle over the cell shows.

We probably will eventually need some UI to prevent or discourage this, e.g. a confirmation dialog, but I think the naive response when this happens should be to clear the zoom row entirely.

Even if the mouse itself isn't the one that got deleted, the shifting of indices may still cause the zoom row to be showing the wrong mouse.

src/models/spaces/cell-zoom/cell-zoom.ts
sfentress: It seems slightly cumbersome to me too. Would it work to simply add the backpack reference directly to to the `CellMouseModel`? Then you could just have an array of two `CellMouseModel` that know who their backpack mice are.

When you add a new mouse, you could check the existing `CellMouseModel` to see if it has the same reference.

I may be missing something that makes this not as workable.

Note: In either case (storing a reference or doing what you're doing now) you'll still have index-changing issues when the user deletes any mouse.

src/models/stores.ts
```diff
@@ -29,17 +30,13 @@ export function createStores(params: ICreateStores, authoring: any): IStores {
...
-           organisms: {
-            one: CellMouseModel.create({ id: "one", baseColor: "brown" }),
-            two: CellMouseModel.create({ id: "two", baseColor: "white" })
-          },
-          rows: [
-            CellZoomRowModel.create({organism: "one"}),
-            CellZoomRowModel.create({organism: "two"}),
-          ]
-         })
+    backpack: params && params.backpack
```
sfentress: I don't think this is needed now that the populations-backpack feature is in, as the user can now add their own mice to the backpack.

src/models/spaces/cell-zoom/cell-zoom.ts
sfentress: (I guess GitHub didn't treat this as part of the review because it's a response, so it's out of order. "The index-changing issues" refers to a comment below.)

ekosmin:
Thanks so much for the review @sfentress! I changed the state like you suggested and everything works *much* more nicely.

Now, `cell-zoom` keeps track of `cell-mice` as before, except there are only ever a maximum of 2. Every `cell-mouse` keeps a reference to its `backpack-mouse` (which I renamed for clarity). We don't store any references to backpack indices now, so the code handles now handles deleting the first mouse if the second mouse is being inspected (deleting an active mouse is still bad but should be an easy fix later). Equality checks are done on mouse objects now instead of indices which reads much more cleanly.

This is rebased on master so it's good to merge if it looks alright to you.

src/models/stores.ts
ekosmin: I figured we might want the ability to show default mice at some point, so I left it for now

src/components/collect-button.tsx
ekosmin: Made this change, much cleaner!

src/components/collect-button.tsx
ekosmin: ^ agreed, that would look nicer, I'll give it a try if I have a chance to do some cleanup

src/models/spaces/cell-zoom/cell-zoom.ts
```diff
@@ -83,46 +83,33 @@ export const kOrganelleInfo: OrganelleInfo = {
...
+      clearRowBackpackMouse,
+      setRowBackpackMouse(rowIndex: number, backpackMouse: BackpackMouseType) {
+        let cellMouse = self.cellOrganisms.find((existingCellMouse) => {
+          return existingCellMouse.backpackMouse === backpackMouse;
         });
+        if (!cellMouse) {
+          cellMouse = CellMouseModel.create({ backpackMouse });
+          self.cellOrganisms.push(cellMouse);
+        }
+        self.rows[rowIndex] = CellZoomRowModel.create({ cellMouse });
```
ekosmin: Here's the new logic, much cleaner - if there's already a `cellMouse` for the given `backpackMouse` we use that. When we delete a mouse, we check if any row is still referencing its cell mouse, and if not we delete it.

Would be nice to have some unit tests for this but I want it in before I disappear for a bit

src/models/stores.ts
sfentress: But if we merge this in there will be no way to open a model without the two collected mice.

sfentress:
@mklewandowski There were a number of conflicts when rebasing around the collect-buttons. Would you mind giving this a once-over to make sure I didn't miss anything? I made a couple decisions regarding the look of the collect-button when used in the corner of the cell-zoom (e.g. the mouse does not have a hover state, since clicking it does nothing; the X button does have a hover state, but only when it itself is hovered, not the mouse). Feel free to suggest changes if you think anything looks off about it.

You can test the behavior and styles at http://connected-bio-spaces.concord.org/branch/backpack-selection/

mklewandowski:
It's looking good and all of my changes seem to be intact (from what I can see).  A couple things:
- I agree that it's a little harder to test without being able to add the test mice, but I think that's fine and something that we just need to deal with.
- we will probably want Michael's opinion on the look of the collect button in cell-zoom corner.  I think the decisions that you made are fine for now, but I bet Michael will be able to suggest some improvements that we can handle in a later story/branch.

Looks good!

PR #20, opened 2018-11-29
Update left nav panel based on new UI specs

mklewandowski:
Update left nav panel based on new UI specs - this includes the new font sizes and faces, hover states, button icons. positions, and colors.


mklewandowski:
I wasn't sure on the other icons.  Originally I thought they were worth keeping in case we needed to add a button here or an icon there during development (and hence we had a standard set to pull from).  What do others think?  Is it time to yank them?

mklewandowski:
Keep in mind too that the DNA/protein view doesn't really have any styling yet.  I've added some colors and an icons to sync it up with the others, but we'll need to get styling if we ever decide to add that space (maybe not now, but perhaps in the future).

src/components/collect-button.sass
```diff
@@ -64,15 +78,26 @@
...
+
+.collect-button-holder:hover .x-close
+  background-color: $color-nav-cerulean-dark1
 
-  .collect-button:hover
-    border-color: $color-nav-cerulean-dark2
-    cursor: pointer
\ No newline at end of file
+.collect-button-holder:hover .x-close.uncollected
+  background-color: transparent
```
sfentress: @mlewandowski Random question (trying to merge in Eli's branch):
is this
```
.collect-button-holder:hover .x-close.uncollected
  background-color: transparent
```
just to clear the prior effects of highlighting `.collect-button-holder:hover .x-close` when it's unselected? Or is it actually important that it is transparent?

If it's the former, could we also just change lines 99-100 to be
```
.collect-button-holder:hover .x-close:not(.uncollected)
  background-color: $color-nav-cerulean-dark1
```

Or, alternatively, could we just not render the `.x-close` in the empty button at all? Is there any reason for it?

mklewandowski: I know you already got to this, but, yes, everything you said is true.  Not rendering the x-close is best.

PR #21, opened 2018-11-29
Zoom control

dstrawberrygirl:
Simple zoom controls moved from the two-up to the Organism Container, which can then choose to show organism, cell, or other views. This PR could be merged in to another if that makes more sense, but I felt it might help highlight the changes I made so it could be used by others.


src/components/spaces/cell-zoom/organism-container.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { ZoomControl } from "../../zoom-control";
+import { OrganismView } from "./organism-view";
+
+import { CellZoomComponent } from "../cell-zoom/cell-zoom";
+
+interface IProps extends IBaseProps {
+  rowIndex: number;
+}
+interface IState {
+  zoomLevel: number;
```
sfentress: @dstrawberrygirl Do you this is would work to integrate this with the `CellZoomRowModel.zoomLevel` property?

sfentress:
This looks great. The only remaining question for me is whether there is any purpose to the `CellZoomComponent` anymore, now that it's nothing more than a thin wrapper around `OrganelleWrapper`. Unless maybe the assay buttons will go there? But they would probably make more sense to go in `OrganismContainer`. In any case, it feels like we have one too many nested boxes.

Maybe @ekosmin has thoughts on this?

That aside, this looks great to me.

dstrawberrygirl:
I agree on the presence of the cellZoom now that I've moved the other pieces out. Aside from assay pieces, we don't yet have any animated pieces for the zoom - I wouldn't want animation code mixed in with cell model code, but equally if we need animation between pieces then we could add it later.

ekosmin:
I agree the assay buttons can go on the graph side - we're already out of room on the cell side. That being said, I agree the `CellZoomComponent` can go

dstrawberrygirl:
Following a Slack conversation regarding the uncertainty over how we'll be addressing development for Assay tools, team have decided to leave the mostly-empty `CellZoomComponent` intact for now.

PR #22, opened 2018-12-04
Zoom cleanup

ekosmin:
After more conversation about the UI, I decided it *was* ultimately best to remove the `CellZoomView`. When I went to make the change, I realized I had made an error in naming conventions...I had been referring to the "Organisms" space as the "Cell Zoom" space throughout the code. Chris's code correctly refers to an "Organisms" space, so I wanted to update the rest of the code to match. Now, all models, components, folders and variables should refer to "Organisms" rows, containers, mice, etc.

This ended up being a big change, and it might actually not be mergeable depending on the PRs people have outstanding. Let me know what you think of the idea behind this change at the very least, and we can talk about merging tomorrow.

In general, the "payoff" here is that the Organism and Population views now have similar naming conventions: in the components, we have an `organisms-space` which displays `organisms-component`s next to graphs just as `populations-space` shows a `populations-component` next to a graph.


ekosmin:
The main question I have is about the use of plural "Organisms", which I used because it's consistent with "Populations". Reading through the code, I'm realizing "Organism" might be clearer - it's possible neither should be pluralized. I figure "Populations" looks that way because of `Populations.js`, but that seems more like an implementation detail.

Don't want to spend much more time on this though, let's catch up tomorrow morning.

PR #24, opened 2018-12-05
Add ability to author markdown space instructions from authoring tool

mklewandowski:


src/components/authoring.tsx
```diff
@@ -168,14 +183,21 @@ export const defaultAuthoring = {
         tan: 33.33
       }
     }
+  },
+  organism: {
```
mklewandowski: @sfentress We're going to need to work with Michael to determine how to show the Organism level instructions.  For now, they exist in the authoring tool, but they will not be placed anywhere until we settle on how the instructions would be displayed in the 4-up.

src/components/instructions.sass
```diff
@@ -0,0 +1,6 @@
+@import vars
+
+.authored-markdown
+  font-family: $font-stack
+  padding: $component-padding
+
```
mklewandowski: @sfentress This sass file is sparse for now, but I wanted it separated because I assume we will have to add more formatting to the authored block once PIs have some time to play with the feature.

src/components/two-up-display.sass
```diff
@@ -48,6 +48,9 @@
     overflow: hidden
     position: relative
 
+  .scrollable
+    overflow: auto
+
```
mklewandowski: @sfentress I wasn't sure if making the entire `content` class scrollable would break how you display the Populations model where overflow is hidden.  For now, I'm only adding the `scrollable` class to the right panel, but even that might need to be more nuanced (in case it clashes with how the graph is displayed.

sfentress: Yeah, we'll want to keep an eye on it, because of all the parts that expect to know their container's size, but it seems to work fine right now. If you add enough text to force the scroll bar to show, it does, and the graph still seems to size correctly. :+1:

PR #26, opened 2018-12-06
Zoom to proteins

sfentress:
This adds the cell model revisions, including the new two-protein version of the zoomed-in model.

This also adds scaling for the model. Scaling is only applies on the first render, as the Organelle model doesn't know how to resize itself.


src/components/spaces/organisms/organelle-wrapper.tsx
```diff
@@ -231,13 +226,44 @@ export class OrganelleWrapper extends BaseComponent<OrganelleWrapperProps, Organ
...
-        <div id={this.props.elementName} className="model" onMouseLeave={this.resetHoveredOrganelle}/>
+        <div id={this.props.elementName} className="model" onMouseLeave={this.resetHoveredOrganelle}
+          ref={this.setModelDomRef} />
         {hoverDiv}
         {droppers}
       </div>
     );
   }
 
+  private createNewModel() {
```
ekosmin: I was having trouble getting new cells to render before without completely removing the old model first...looks like this is the logic I needed to make that work!

src/components/spaces/organisms/organelle-wrapper.tsx
```diff
@@ -231,13 +226,44 @@ export class OrganelleWrapper extends BaseComponent<OrganelleWrapperProps, Organ
...
+    if (this.model) {
+      this.model.destroy();
+      if (this.modelDomRef) {
+        ((this.modelDomRef as unknown) as Element).innerHTML = "";
+      }
+    }
+    const {zoomLevel } = this.props;
```
ekosmin: Missing a space here

src/components/spaces/organisms/organelle-wrapper.tsx
```diff
@@ -114,19 +114,33 @@ export class OrganelleWrapper extends BaseComponent<OrganelleWrapperProps, Organ
...
   public updateReceptorImage() {
     const model = this.getModel();
-    if (model.world.getProperty("working_receptor")) {
-      model.view.hide("#receptor-broken", true);
-      if (model.world.getProperty("hormone_bound")) {
-        model.view.hide("#receptor-working", true);
-        model.view.show("#receptor-bound", true);
+    const { zoomLevel } = this.props;
+    if (zoomLevel === "cell") {
+      model.view.show(".receptor-mini", true);
```
ekosmin: I wonder if ultimately we'll want to see the difference between broken and functional proteins at the cell level? Either way, I agree this is good for now since the receptors look the same at that scale anyway.

src/models/spaces/organisms/organisms-mouse.ts
```diff
@@ -32,7 +32,8 @@ export const OrganismsMouseModel = types
           open_gates: false,
           eumelanin: getPercentDarkness(),
           hormone_spawn_period: 40,
-          working_receptor: workingReceptor
+          working_receptor: workingReceptor,
+          tan: self.backpackMouse.baseColor === "tan"
```
ekosmin: Might be a little more readable if there was just a single property for color

sfentress:
@ekosmin I fixed the style issue, and replaced the odd "working_receptor" and "tan" properties with a single prop, which works better. This also fixes the omission that broken g-protein parts should appear at the receptor they broke on.

(The whole logic of needing to split all the organelle rules into "first receptor" and "second receptor" everywhere is annoying, and a limitation of organelle rules, which we can come back to at some point...)

PR #27, opened 2018-12-06
Line chart improvements

dstrawberrygirl:
Added timeline control to populations line chart to make it possible to scroll back in time (while paused) and view all data. Added rc-slider to make it possible to style the slider, plus made some small tweaks to the chart models to make it easier to work with the data.


src/components/main-content.tsx
```diff
@@ -10,16 +10,17 @@ import { OrganismsSpaceComponent } from "./spaces/organisms-space";
...
 
 type SpaceTypeToComponent = {
   [key in SpaceType]: SpaceComponent | undefined
 };
 
 const kSpaceComponents: SpaceTypeToComponent = {
   populations: PopulationsSpaceComponent,
-  breeding: undefined,
   organism: OrganismsSpaceComponent,
+  breeding: undefined,
```
sfentress: Just wondering, were the set of changes on this file requested by tslint? The current version doesn't seem to complain for me, on my editor. Don't know if we have different settings.

src/components/spaces/chart-test-space.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+import { PopulationsComponent } from "./populations/populations-container";
+import { ChartTest } from "../charts/chart-test";
+
+interface IProps extends IBaseProps {}
+interface IState {}
+
+@inject("stores")
+@observer
+export class ChartTestSpaceComponent extends BaseComponent<IProps, IState> {
```
sfentress: Is this component ever used, or accessible?

src/models/spaces/populations/mouse-model/mouse-populations-model.ts
```diff
@@ -11,23 +11,31 @@ const chartData = {
       name: "White mice",
       dataPoints: [],
       color: "#f4ce83",
-      maxPoints: 100,
-      fixedMin: 0,
-      expandOnly: true
+      maxPoints: 20,
+      fixedMinA2: 0,
+      fixedMaxA2: 50,
+      expandOnly: true,
```
sfentress: I think `expandOnly` is ignored when `fixedMaxA2` is present, right?

src/models/spaces/populations/populations.ts
```diff
@@ -33,7 +33,15 @@ export const PopulationsModel = types
...
-      self.isPlaying = self.model.interactive.isPlaying;
+      if (self.isPlaying !== self.model.interactive.isPlaying) {
+        self.isPlaying = self.model.interactive.isPlaying;
+
+        if (self.isPlaying) {
+          // clear any subset information and return to viewing the live tail
+          const d: ChartDataModelType = self.model.chartData;
+          d.setDataSetSubset(-1, d.maxPoints);
+        }
+      }
```
sfentress: Couldn't this be handled by the chart observing its own `isPlaying` property? Since the chart is the one enforcing that a user can't scrub out of the live-tail view when `isPlaying` is true, it seems like this should behavior should be the chart's concern.

src/components/main-content.tsx
dstrawberrygirl: I had a temporary space set up for testing the charts with the chart-test-space - this is a remnant of that testing and shall be removed, thanks!

src/components/spaces/chart-test-space.tsx
dstrawberrygirl: not any more - I'll remove it, thanks

src/models/spaces/populations/mouse-model/mouse-populations-model.ts
dstrawberrygirl: yes - not sure if this is correct behavior or whether we should have the "fixed" max as a baseline and, if expand is enabled, expand as required.

src/components/main-content.tsx
dstrawberrygirl: ah misunderstood - comment was referring to the change in the order of components (a remnant of testing) but has no effect on functionality.

src/models/spaces/populations/populations.ts
dstrawberrygirl: Good idea - have moved this logic to the chart.

PR #30, opened 2018-12-11
Assay tools

ekosmin:
Adds simple assay controls to the cell view, with substance addition controls soon to follow.

### DIFF
--- a/src/components/spaces/organisms/organelle-wrapper.tsx
+++ b/src/components/spaces/organisms/organelle-wrapper.tsx
@@ -362,9 +362,7 @@ export class OrganelleWrapper extends BaseComponent<OrganelleWrapperProps, Organ
     model.on("view.hover.enter", (evt: any) => {
       const {organisms} = this.stores;
       const hoveredOrganelle = this.getOrganelleFromMouseEvent(evt);
-      if (hoveredOrganelle) {
-        organisms.rows[this.props.rowIndex].setHoveredOrganelle(hoveredOrganelle);
-      }
+      organisms.rows[this.props.rowIndex].setHoveredOrganelle(hoveredOrganelle);
     });
 
     model.on("view.click", (evt: any) => {
@@ -452,7 +450,9 @@ export class OrganelleWrapper extends BaseComponent<OrganelleWrapperProps, Organ
   }
 
   private resetHoveredOrganelle() {
-    this.setState({hoveredOrganelle: null});
+    const {organisms} = this.stores;
+    const {rowIndex} = this.props;
+    organisms.rows[rowIndex].setHoveredOrganelle(undefined);
   }
 
   private getOpaqueSelector(organelleType: OrganelleType) {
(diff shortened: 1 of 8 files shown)


src/components/spaces/organisms/organelle-wrapper.tsx
```diff
@@ -265,6 +271,13 @@ export class OrganelleWrapper extends BaseComponent<OrganelleWrapperProps, Organ
     });
   }
 
+  private forceDropper(e: any) {
+    // Hack to force Fabric canvases to inherit cursor styles, should configure in Organelle instead
```
ekosmin: This is the same hack I used in `connected-bio` to keep the dropper from turning back into a pointer...let me know if you think there's a cleaner solution.

src/components/spaces/organisms/organisms-container.tsx
```diff
@@ -90,7 +90,7 @@ export class OrganismsContainer extends BaseComponent<IProps, IState> {
             {
               organismsMouse != null &&
                 <OrganelleWrapper zoomLevel={zoomLevel} elementName={`organelle-wrapper-${rowIndex}`}
-                  rowIndex={rowIndex} width={width}/>
+                  rowIndex={rowIndex} width={width} mode={mode}/>
```
ekosmin: Still not entirely sure why these props need to be passed explicitly like this, but it did the trick!

src/components/spaces/organisms/organelle-wrapper.tsx
```diff
@@ -100,10 +102,22 @@ export class OrganelleWrapper extends BaseComponent<OrganelleWrapperProps, Organ
...
-    const zoomLevel = this.props.zoomLevel;
-    if (this.model && zoomLevel !== this.model.zoomLevel) {
-      this.createNewModel();
+    const { zoomLevel, mode } = this.props;
+    if (this.model) {
+      if (zoomLevel !== this.model.zoomLevel) {
+        this.createNewModel();
+      }
+
+      if (mode !== this.model.mode) {
```
ekosmin: I wasn't sure what the best way to pause the model was so I followed your lead here, let me know what you think.

sfentress: Yeah, I don't know whether there's a better way to do this. I basically wanted the equivalent of `componentDidUpdate(prevProps)` so that we could compare the props before and after, and call these functions on the model, but I couldn't see how MST supports that. So I stuck the `zoomLevel` prop on the model just as a way of caching the value. It could have also just been a `let` outside the class. It's not the right solution, I think, but I couldn't work out the right solution.

src/components/spaces/organisms/organisms-container.tsx
sfentress: Again, there may be something we're not doing right with MST. OrganelleWrapper ought to be able to just bring these things in directly from the model observers, right? But somehow the fact that it's in an array or something means that it doesn't trigger a notification when things change.

PR #31, opened 2018-12-11
161595479 style population controls

mklewandowski:
Updated population controls based on UI specs.  Eventually there will be minor additions to the population control styling, but we don't have mock-ups yet (this includes things like the final button icons).  We can add that content in separate stories and branches once we have them.


cypress/integration/population.test.js
```diff
@@ -30,29 +30,26 @@ context("Test the population level", () => {
...
+      cy.get('[data-test=pop-toolbar]').find('button').eq(0).click({force:true}); //add mice and run
     });
 
-    it("verifies select button functionality", () => {
+    it("verifies collect button functionality", () => {
       cy.get('[data-test=pop-toolbar]').find('button').eq(0).click({force:true});
       cy.get('[data-test=pop-toolbar]').contains('Run');
-      cy.get('[data-test=select-button]').click({force:true});
+      cy.get('[data-test=collect-button]').click({force:true});
       //Need to be able to read iframe data in order to locate mouse coordinates
```
mklewandowski: @kevinsantos-cc I updated the Cypress tests based on the new button arrangement, but you might want to double-check my work.

src/components/spaces/populations/populations-container.tsx
```diff
@@ -28,87 +28,165 @@ export class PopulationsComponent extends BaseComponent<IProps, IState> {
...
-              </div>
-            );
-          }}
-        </SizeMe>
+              );
+            }}
+          </SizeMe>
+          <div className="populations-toolbar" data-test="pop-toolbar">
+            <div className="toolbar-row">
+              <button className={"population-button " + runButtonClass}
```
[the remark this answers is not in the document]
sfentress: :+1: That sounds right to me. If the peas unit doesn't want those buttons, we'll cross that bridge when we come to it.

src/models/spaces/populations/populations.ts
```diff
@@ -12,23 +13,32 @@ export function createPopulationsModel(curriculumName: string, authoring: any):
...
+        interactionMode: "none"
       });
   }
 }
 
 export interface ToolbarButton {
   title: string;
+  imageClass?: string;
+  secondaryTitle?: string;
+  secondaryTitleImageClass?: string;
```
sfentress: This seems like a good way to handle the double-line label.

src/models/spaces/populations/mouse-model/mice.ts
```diff
@@ -68,7 +68,7 @@ export function getMouseSpecies(model: MousePopulationsModelType) {
       this.set("age", Math.round(Math.random() * 10));
 
       if (this.organism) {
-        if (!model["inheritance.breedWithoutInheritance"]) {
+        if (!model["inheritance.breedWithInheritance"]) {
```
sfentress: This boolean needs to be flipped now that the meaning has changed. (i.e. the `!` needs to be removed.)

mklewandowski: Pushed a change to fix this.

mklewandowski: I also rebased into master so we will be ready to merge.

src/components/two-up-display.tsx
```diff
@@ -58,6 +58,7 @@ export class TwoUpDisplayComponent extends BaseComponent<IProps, IState> {
         <div className="content scrollable" data-test="right-content">
           {this.props.rightPanel}
         </div>
+        <div className="footer" data-test="right-footer"/>
```
sfentress: Looking back on this, I'm not certain that these footer elements quite make sense like this, because `rightPanel` can't easily put elements in them. Also I'm not sure that they should show for instructions.

mklewandowski: They are in all panels in the UI spec, but I don't like them.  They take up space and provide no functionality.  They seem to only exist for aesthetics.

sfentress: Ok, yeah, I think they're supposed to eventually have UI in them, but until then they are taking up space. So I think I will remove them in my branch until we've added functionality to them.

PR #33, opened 2018-12-12
Amino acid view, and right-panel buttons

sfentress:
This adds the protein viewer to the organism space. Along the way I had to add the three buttons, so we could access it, apologies if that's confusing and ought to have been a different PR.

1. Add buttons to right hand panel, and string enum modes to right panel in `ui`, to replace the `showGraph` booleans
2. Add re-written protein viewer code from https://github.com/concord-consortium/protein-viewer
3. Add `inspect` mode to cell space, as well as `selectedOrganelle`
4. Display the selected protein in the top right view


src/components/spaces/organisms-space.tsx
```diff
@@ -45,10 +45,18 @@ export class OrganismsSpaceComponent extends BaseComponent<IProps, IState> {
       row.setShowProteinDNA(!row.showProteinDNA);
   }
 
+  private handleSetSelectStartPercent = (percent: number) => {
+    const { organisms } = this.stores;
+    // set on both simultaneously
+    organisms.rows[0].setProteinSliderStartPercent(percent);
+    organisms.rows[1].setProteinSliderStartPercent(percent);
```
ekosmin: Nice example of synchronization here! I think I see why new mice don't match position immediately - when a new mouse is added, a fresh row is created and wipes the position set here.

src/models/ui.ts
```diff
@@ -12,7 +15,8 @@ export function createUIModel(authoring: any): UIModelType {
         displayedSpace = "none";
   }
   return UIModel.create({
-    showOrganismGraph: [false, true],
+    populationsRightPanel: "instructions",
+    organismRightPanel: ["instructions", "data"],
```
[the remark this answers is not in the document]
sfentress: This is now changed in #38

src/components/spaces/organisms-space.tsx
sfentress: This ought to be fixed now in #38

PR #34, opened 2018-12-13
Cleaning up the assay/organism view

ekosmin:
Small style changes if you don't mind taking a quick look @mklewandowski!

- "Assay" button now reads "Measure"
- Hovered organelle is moved to the top right corner for smaller screens
- Organism label is hidden

### DIFF
--- a/src/components/spaces/organisms/organelle-wrapper.sass
+++ b/src/components/spaces/organisms/organelle-wrapper.sass
@@ -10,11 +10,11 @@
 
 .hover-location
     position: absolute
-    bottom: 5px
+    top: 45px
     right: 5px
     min-width: 94px
     height: 18px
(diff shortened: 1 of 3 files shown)


src/components/spaces/organisms/organelle-wrapper.sass
```diff
@@ -10,11 +10,11 @@
...
 .hover-location
     position: absolute
-    bottom: 5px
+    top: 45px
     right: 5px
     min-width: 94px
     height: 18px
-    background: #FFFFFF88
+    background: #FFFFFFAA
     border: 1px solid #444
```
mklewandowski: This reminds me that at some point we should remember to clean up all rogue colors and only use colors from the spec - but that can wait for another day.

PR #36, opened 2018-12-13
162487966 multiple assays

mklewandowski:
This branch includes changes to allow for multiple organism level assays to be displayed in a graph in the right panel which follows the layout/format of the previous version of connected bio.  Graphs are scaled to fit into a given view (which includes changes to allow the population graph to do so as well).  A few notes:
 - there is probably some cleanup needed here, but I would rather have people start looking at it sooner rather than later
- Editing my previous comment.  I squashed the commit history down to just two items: the changes to add multiple assays to the graph and the changes to improve graph sizing and scaling.  This cleans up some of the wildness from yesterday.  These commits are not as nuanced as I would like, but it's better than the 10 or so previous commits that had changes being added and then removed.
- this has been rebased into master, and I made some changes to get it working with the last batch of changes there.  Nonetheless, changes are coming in rapidly, so we should double-check to make certain content works as expected in all views

### DIFF
--- a/src/components/charts/bar-chart.tsx
+++ b/src/components/charts/bar-chart.tsx
@@ -15,7 +15,7 @@ interface IBarProps {
 
 const defaultOptions: ChartOptions = {
   title: {
-    display: true,
+    display: false,
     text: "",
     fontSize: 22
   },
@@ -82,11 +82,12 @@ const barData = (chartData: ChartDataModelType) => {
       dset.backgroundColor = ChartColors.map(c => hexToRGBValue(c.hex, seriesOpacity));
       dset.borderColor = ChartColors.map(c => hexToRGBValue(c.hex, 1.0));
     }
+    dset.stack = d.stack;
     barDatasets.push(dset);
   }
 
   const barChartData = {
-    labels: chartData.dataLabels,
+    labels: chartData.chartLabels,
     datasets: barDatasets
   };
 
--- a/src/components/charts/line-chart.tsx
+++ b/src/components/charts/line-chart.tsx
@@ -30,7 +30,7 @@ const defaultOptions: ChartOptions = {
     display: true,
     position: "bottom",
   },
-  maintainAspectRatio: true,
+  maintainAspectRatio: false,
   scales: {
     display: false,
     yAxes: [{
@@ -161,8 +161,10 @@ export class LineChart extends BaseComponent<ILineProps, ILineState> {
     );
 
     return (
-      <div className="line-chart-container" data-test="line-chart">
-        {graphs}
+      <div className="line-chart-container">
+        <div className="line-chart-container" data-test="line-chart">
+          {graphs}
+        </div>
         <LineChartControls chartData={chartData} isPlaying={isPlaying} />
       </div>
     );
--- a/src/components/two-up-display.tsx
+++ b/src/components/two-up-display.tsx
@@ -43,7 +43,7 @@ export class TwoUpDisplayComponent extends BaseComponent<IProps, IState> {
         <div className={"header " + this.props.spaceClass} data-test="left-header">
           <div className="title" data-test="left-title">{this.props.leftTitle}</div>
         </div>
-        <div className="content" data-test="left-content">
+        <div className={"content " + this.props.spaceClass} data-test="left-content">
           {this.props.leftPanel}
         </div>
       </div>
@@ -67,7 +67,9 @@ export class TwoUpDisplayComponent extends BaseComponent<IProps, IState> {
   private renderRightPanel() {
     const { selectedRightPanel, instructionsIconEnabled, dataIconEnabled, informationIconEnabled } = this.props;
     const title = titles[selectedRightPanel];
-    const contentClass = "content" + (selectedRightPanel === "instructions" ? " scrollable" : "");
+    const contentClass = "content " + (selectedRightPanel === "instructions" ? " scrollable " : "")
+                                    + (selectedRightPanel === "data" ? "resizetofit " : "")
+                                    + this.props.spaceClass;
     return (
       <div className="two-up-panel right-abutment">
         <div className={"header " + this.props.spaceClass} data-test="right-header">
--- a/src/models/spaces/charts/chart-data.ts
+++ b/src/models/spaces/charts/chart-data.ts
@@ -4,9 +4,15 @@ import { ChartDataSetModel, ChartDataSetModelType, ChartColors } from "./chart-d
 export const ChartDataModel = types
   .model("ChartData", {
     name: types.string,
-    dataSets: types.array(ChartDataSetModel)
+    dataSets: types.array(ChartDataSetModel),
+    labels: types.array(types.string)
   })
   .views(self => ({
+    get chartLabels() {
+      if (self.labels && self.labels.length > 0) {
+        return self.labels;
+      } else return [];
+    },
     // labels for a data point - essential for a bar graph, optional for a line
     get dataLabels() {
       if (self.dataSets && self.dataSets.length > 0) {
--- a/src/models/spaces/organisms/organisms-row.ts
+++ b/src/models/spaces/organisms/organisms-row.ts
@@ -1,8 +1,8 @@
 import { types, Instance } from "mobx-state-tree";
 import { ChartDataModelType, ChartDataModel } from "../charts/chart-data";
-import { DataPoint, ChartDataSetModel, ChartColors, DataPointType } from "../charts/chart-data-set";
+import { DataPoint, ChartDataSetModel, ChartDataSetModelType, DataPointType } from "../charts/chart-data-set";
 import { OrganismsMouseModel } from "./organisms-mouse";
-import { kSubstanceNames } from "./organisms-space";
+import { kSubstanceInfo, kSubstanceNames } from "./organisms-space";
 
 export const Organelle = types.enumeration("type", [
   "nucleus",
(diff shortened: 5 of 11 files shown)


src/components/charts/bar-chart.tsx
```diff
@@ -15,7 +15,7 @@ interface IBarProps {
 
 const defaultOptions: ChartOptions = {
   title: {
-    display: true,
+    display: false,
```
mklewandowski: I assume we no longer want to show the title since the panel header should act as a title.

src/components/charts/line-chart.tsx
```diff
@@ -30,7 +30,7 @@ const defaultOptions: ChartOptions = {
     display: true,
     position: "bottom",
   },
-  maintainAspectRatio: true,
+  maintainAspectRatio: false,
```
mklewandowski: Allows the populations graph to fill the panel space much better.

src/models/spaces/charts/chart-data.ts
```diff
@@ -4,9 +4,15 @@ import { ChartDataSetModel, ChartDataSetModelType, ChartColors } from "./chart-d
 export const ChartDataModel = types
   .model("ChartData", {
     name: types.string,
-    dataSets: types.array(ChartDataSetModel)
+    dataSets: types.array(ChartDataSetModel),
+    labels: types.array(types.string)
   })
   .views(self => ({
+    get chartLabels() {
```
mklewandowski: These are the chart level labels as opposed to the labels on the datasets (this is where we store the assayed region for future display in the chart)

src/models/spaces/organisms/organisms-row.ts
```diff
@@ -30,36 +30,52 @@ export const OrganismsRowModel = types
     hoveredOrganelle: types.maybe(Organelle),
     selectedOrganelle: types.maybe(Organelle),
     mode: types.optional(Mode, "normal"),
-    assayedOrganelle: types.maybe(Organelle),
+    assayedOrganelles: types.array(Organelle),
```
mklewandowski: I'm building up a simple array of assayed regions, but thinking back to a comment in the meeting yesterday do we need to think about ways to clear/reset this?

src/models/spaces/organisms/organisms-row.ts
```diff
@@ -30,36 +30,52 @@ export const OrganismsRowModel = types
...
     proteinSliderStartPercent: 0
   })
   .views(self => ({
     get currentData(): ChartDataModelType {
-      const chartDataSets = [];
-      const points: DataPointType[] = [];
-      const organelle = self.assayedOrganelle;
-      if (organelle) {
-        kSubstanceNames.forEach((substance) => {
+      const chartDataSets: any[] = [];
```
mklewandowski: There might be some obvious cleanup in the chunk.  I'm open to thoughts.

src/models/spaces/organisms/organisms-row.ts
```diff
@@ -30,36 +30,52 @@ export const OrganismsRowModel = types
...
-        dataPoints: points,
-        pointColors: ["#f4ce83", "#d88bff", "#795423", "#0adbd7"]
-      }));
+        const stackNum = "Stack " + stackIterator;
+        chartDataSets.push(ChartDataSetModel.create({
+          name: substanceName,
+          dataPoints: points,
+          color: substanceColor,
+          fixedMaxA2: 800,
+          fixedMaxA1: 800,
```
mklewandowski: I set the max to 800 which was used in the previous version.

src/components/two-up-display.tsx
```diff
@@ -67,7 +67,9 @@ export class TwoUpDisplayComponent extends BaseComponent<IProps, IState> {
   private renderRightPanel() {
     const { selectedRightPanel, instructionsIconEnabled, dataIconEnabled, informationIconEnabled } = this.props;
     const title = titles[selectedRightPanel];
-    const contentClass = "content" + (selectedRightPanel === "instructions" ? " scrollable" : "");
+    const contentClass = "content " + (selectedRightPanel === "instructions" ? " scrollable " : "")
+                                    + (selectedRightPanel === "data" ? "resizetofit " : "")
+                                    + this.props.spaceClass;
```
mklewandowski: These two-up panels are having to handle more and more display cases.  We need to keep our eye on the CSS here to ensure everything displays as expected.

src/models/spaces/organisms/organisms-row.ts
```diff
@@ -30,36 +30,52 @@ export const OrganismsRowModel = types
...
     showProteinAminoAcidsOnProtein: false,
     proteinSliderStartPercent: 0
   })
   .views(self => ({
     get currentData(): ChartDataModelType {
-      const chartDataSets = [];
-      const points: DataPointType[] = [];
-      const organelle = self.assayedOrganelle;
-      if (organelle) {
-        kSubstanceNames.forEach((substance) => {
```
mklewandowski: There might be some obvious cleanup in the chunk.  I'm open to thoughts.

src/models/spaces/charts/chart-data.ts
```diff
@@ -4,9 +4,15 @@ import { ChartDataSetModel, ChartDataSetModelType, ChartColors } from "./chart-d
 export const ChartDataModel = types
   .model("ChartData", {
     name: types.string,
-    dataSets: types.array(ChartDataSetModel)
+    dataSets: types.array(ChartDataSetModel),
+    labels: types.array(types.string)
   })
   .views(self => ({
```
mklewandowski: These are the chart level labels as opposed to the labels on the datasets (this is where we store the assayed region for future display in the chart)

src/models/spaces/charts/chart-data.ts
```diff
@@ -4,9 +4,15 @@ import { ChartDataSetModel, ChartDataSetModelType, ChartColors } from "./chart-d
...
+    get chartLabels() {
+      if (self.labels && self.labels.length > 0) {
+        return self.labels;
+      } else return [];
+    },
     // labels for a data point - essential for a bar graph, optional for a line
```
mklewandowski: These are the chart level labels as opposed to the labels on the datasets (this is where we store the assayed region for future display in the chart)

src/models/spaces/charts/chart-data.ts
```diff
@@ -4,9 +4,15 @@ import { ChartDataSetModel, ChartDataSetModelType, ChartColors } from "./chart-d
...
-    dataSets: types.array(ChartDataSetModel)
+    dataSets: types.array(ChartDataSetModel),
+    labels: types.array(types.string)
   })
   .views(self => ({
+    get chartLabels() {
+      if (self.labels && self.labels.length > 0) {
+        return self.labels;
+      } else return [];
+    },
```
mklewandowski: These are the chart level labels as opposed to the labels on the datasets (this is where we store the assayed region for future display in the chart)

src/models/spaces/organisms/organisms-row.ts
```diff
@@ -30,36 +30,52 @@ export const OrganismsRowModel = types
...
+      const colorMaps: any = { pheomelanin: "#f4ce83",
+                               signalProtein: "#d88bff",
+                               eumelanin: "#795423",
+                               hormone: "#0adbd7" };
+      let stackIterator = 0;
+      kSubstanceNames.forEach((substance) => {
+        const substanceName = substance === "signalProtein" ? "signal protein" : substance;
+        const substanceColor: string = colorMaps[substance];
+        const points: DataPointType[] = [];
+        self.assayedOrganelles.forEach((organelle) => {
```
mklewandowski: There might be some obvious cleanup in the chunk.  I'm open to thoughts.

src/models/spaces/organisms/organisms-row.ts
ekosmin: I think this could be type `ChartDataSetModelType[]`

src/models/spaces/organisms/organisms-row.ts
```diff
@@ -30,36 +30,52 @@ export const OrganismsRowModel = types
...
-      const organelle = self.assayedOrganelle;
-      if (organelle) {
-        kSubstanceNames.forEach((substance) => {
+      const chartDataSets: any[] = [];
+      const organelleLabels: string[] = [];
+
+      self.assayedOrganelles.forEach((organelle) => {
+        organelleLabels.push(organelle);
+      });
+      const colorMaps: any = { pheomelanin: "#f4ce83",
```
ekosmin: Just learned you could do this inline in Sam's PR, I'd try a type def like:

```
const colorMaps: {[sub in SubstanceType]: string} = {...
```

Then, if we add any substances in the future, this map will automatically complain if it doesn't have a color associated with it.

src/models/spaces/organisms/organisms-row.ts
ekosmin: Looks good to me!

src/models/spaces/charts/chart-data.ts
mklewandowski: Oddly, I was originally just returning `return self.labels` but mobx was throwing errors left and right about accessing an undefined array.  This explicit check and return of `[]` fixed it.

src/models/spaces/organisms/organisms-row.ts
mklewandowski: Oh, yeah!  Using the `index` in the `forEach` is an obvious one.  Thanks!

src/models/spaces/organisms/organisms-row.ts
mklewandowski: Yup.  Will update.

src/models/spaces/organisms/organisms-row.ts
mklewandowski: 👍

src/models/spaces/charts/chart-data.ts
mklewandowski: 'data.labels' does the same thing.  We get the following error:
`read an array index (0) that is out of bounds (0). Please check length first. Out of bound indices will not be tracked by MobX`

src/models/spaces/organisms/organisms-row.ts
mklewandowski: `kSubstanceInfo` is a good idea.  Will update to include.

src/models/spaces/organisms/organisms-row.ts
mklewandowski: I yanked `colorMaps` entirely now that I'm getting colors out of the `kSubstanceType` that I created.

PR #38, opened 2018-12-13
Amino acid viewer fixes

sfentress:
This is pretty small, except for a shift from storing the right panel state from the ui model to the individual space models.

### DIFF
--- a/src/components/spaces/organisms/organelle-wrapper.tsx
+++ b/src/components/spaces/organisms/organelle-wrapper.tsx
@@ -13,7 +13,6 @@ import { BaseComponent } from "../../base";
 // import { SubstanceType } from "../models/Substance";
 import "./organelle-wrapper.sass";
 import { OrganelleType, ModeType, ZoomLevelType } from "../../../models/spaces/organisms/organisms-row.js";
-import { IProps, IState } from "populations-react/build/lib/components/PopulationsView";
 
 interface OrganelleWrapperProps {
   zoomLevel: ZoomLevelType;
@@ -135,7 +134,7 @@ export class OrganelleWrapper extends BaseComponent<OrganelleWrapperProps, Organ
   }
 
   public organelleClick(organelleType: OrganelleType, location: {x: number, y: number}) {
-    const { organisms, ui } = this.stores;
+    const { organisms } = this.stores;
     const row = organisms.rows[this.props.rowIndex];
     if (row.mode === "assay") {
       row.setActiveAssay(organelleType);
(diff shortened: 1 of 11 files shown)


src/components/two-up-display.sass
```diff
@@ -64,12 +71,7 @@
           .button
             fill: $two-up-button-icon-color-active
 
-      &.disabled
-        .button
-          opacity: 0.25
-      &.active
-        .button
-          fill: $two-up-button-icon-color-active
+
```
ekosmin: You could unindent and make a slight tweak to the previous version to just define the fill once, right? Something like

```
.button-holder
  &.disabled
    .button
      opacity: 0.25
  &.active
    .button
      fill: $two-up-button-icon-color-active
```

src/models/spaces/organisms/organisms-space.ts
```diff
@@ -123,7 +123,15 @@ export const OrganismsSpaceModel = types
           organismsMouse = OrganismsMouseModel.create({ backpackMouse });
           self.organismsMice.push(organismsMouse);
         }
-        self.rows[rowIndex] = OrganismsRowModel.create({ organismsMouse });
+
+        // sync initial slider position
+        let proteinSliderStartPercent = 0;
```
ekosmin: Now that I think about this, would it be better to store this value in the `organisms-state` model instead? It doesn't seem like there's a time when rows will have different values

src/models/spaces/organisms/organisms-space.ts
```diff
@@ -124,14 +125,20 @@ export const OrganismsSpaceModel = types
           self.organismsMice.push(organismsMouse);
         }
 
+        // keep right panel as it was
```
ekosmin: I see what you mean, recreating the rows each time a mouse is created/destroyed has caused a couple of problems so far. Maybe that was a mistake to begin with, even though it previously simplified resetting.

src/components/spaces/organisms/organelle-wrapper.tsx
```diff
@@ -138,6 +138,7 @@ export class OrganelleWrapper extends BaseComponent<OrganelleWrapperProps, Organ
     const row = organisms.rows[this.props.rowIndex];
     if (row.mode === "assay") {
       row.setActiveAssay(organelleType);
+      row.setRightPanel("data");        // auto-switch to data
```
ekosmin: Rad, thanks Sam!

src/models/spaces/organisms/organisms-space.ts
sfentress: We're synching because I figured that that makes the most sense compared with the old model, but I don't know that it will always be like that. I could imagine at some point students might even select two different kinds of proteins, so then it wouldn't make sense to sync.

If we were always definitely synching, I think putting it there would make sense.

src/components/two-up-display.sass
sfentress: It's already defined exactly like that a few lines up. The issue is that the organism-specific rule

`.header.organism .button-holder .button`

is higher priority than the generic header + active:

`.header .button-holder .button.active`

and so the top one still applies even when the button is active, so I need to create another rule for 

`.header.organism .button-holder .button.active`

If you can get another way to work that's cleaner, go for it, but I didn't see anything that simplified the number of rules.

PR #39, opened 2018-12-14
Update UI based on UI spec

mklewandowski:
A series of small UI updates to match Michael's UI spec.  This includes the following:
-add SVG definitions for icons
-add x close icon for collect button
-update collect buttons x close and index positions
-update collect buttons x close and index colors
-add padding in right panel when displaying instructions
-add correct panel spacing in 4-up display
-add correct hover effect on collect button x close and collect circle button
-add correct collect button state when placing an organism in the organism view
-update collect button x close background shape when hovering over button
-unify transition times


src/index.html
```diff
@@ -158,6 +158,45 @@
       <path d="M17.48,10.42c2.81-1.35,5.08-4.11,8.6-2.72c-1.21,1.08-2.43,2.2-3.7,3.26c-0.55,0.46-1.48,0.81-2.33,0.07
         c-0.29,0.09-0.5,0.16-0.65,0.19C18.63,11.38,18,11.26,17.48,10.42z"/>
       </symbol>
+      <symbol id="icon-collect-cursor" viewBox="0 0 25 20">
```
sfentress: Is this for future use?

mklewandowski: Yes.  There are two population cursors that we need to use at some point.  A collect and an inspect cursor.

PR #40, opened 2018-12-14
Show graph when user select assay tool

mklewandowski:
This change shows the graph when the user selects the assay tool.  This should have been handled in my previous batch of graph updates, but the change was missed.  I was reminded of it when reviewing my notes and wanted to get something in since Frieda specifically requested it during the demo meeting.


mklewandowski:
Great!  Thanks!

sfentress:
Note, the PT story did say to do it when we entered assay mode, but I felt like this UI made more sense, and in any case we would need to unify the behavior with the inspect tool, which switches on the actual inspect. If they want it different, we can easily change it.

PR #42, opened 2018-12-17
Fix double populations bug

sfentress:


src/components/spaces/populations/populations-container.tsx
```diff
@@ -30,6 +30,13 @@ export class PopulationsComponent extends BaseComponent<IProps, IState> {
     super(props);
   }
 
+  public componentWillUnmount() {
+    const populations = this.props.stores && this.props.stores.populations;
```
ekosmin: I wouldn't have thought to include null checks here, were there problems without them?

sfentress: Simplified in 6754961d8cc. Thanks for noticing! (Still confused, though.)

PR #48, opened 2019-02-05
Add population inspect and collect cursors

mklewandowski:
Small fix that came to my attention while reviewing existing in-app assets - add the special inspect and collect cursors in the population simulation.  Colors and sizes are from the UI spec.


mklewandowski:
Michael wants to tweak the icon images.  I'll add on another branch to clarify.

PR #51, opened 2019-02-06
Save student backpack and current space

sfentress:
This

1. Serializes the user's backpack mice and current space every time they change
2. Communicates with LARA using iFramePhone to save the data, and to receive the data when initializing the app
3. Creates the backpack and current space using the saved data, if available

This also tweaks some authoring schema names to make them more directly 1:1 with the MST models.

This is PR 1 of 2 related to state saving. The second was spun off as it has some extra complications.

https://www.pivotaltracker.com/story/show/163457123
https://www.pivotaltracker.com/story/show/163457128

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8637,6 +8637,11 @@
       "resolved": "https://registry.npmjs.org/iferr/-/iferr-0.1.5.tgz",
       "integrity": "sha1-xg7taebY/bazEEofy8ocGS3FtQE="
     },
+    "iframe-phone": {
+      "version": "1.2.0",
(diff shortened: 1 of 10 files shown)


sfentress:
It would make sense for the Populations right panel to retain its state, since that's what happens for the Organims space in #52 . That would be the only state I'd keep, though, since the Populations model should be reset each time.

I don't think it's worth putting in here for the moment though. The student data is expandable, and adding a Populations state would be backwards compatible with any saved data if we add it later.

mklewandowski:
I agree.  I was happy to see the right panel state restoring in the organism level view.  We should probably at least add a story to restore the pop sim right panel so it can be addressed in the future.

PR #52, opened 2019-02-06
Save organisms space state

sfentress:
This saves the current state of the organisms space to the student data, which was set up in #51. This is PR 2 of 2 of state-saving, and was spun off to focus on the extra wrinkles related to saving and restoring the organisms space. This PR is just from 75599e3 (Refactor OrganismsMouse with private substance deltas) onwards.

This serializes and deserializes the organisms model, after first refactoring the `OrganismsMouse` model to remove `substanceDeltas` from the model, which both removes transient props from the model and allows us to avoid `afterCreate`, which was causing problems on model creation. Further, before we can create OrgamismsMice, we need to replace the `BackpackMouse` references, which are serialized as a string id, with actual `BackpackMouse` models, as the `Organisms` tree can't find objects in the `Backpack` tree.


sfentress:
Added authoring type data, which is not really necessary for this PR but starts to help solidify the various types (`ConnectedBioAuthoring`, `UserSaveDataType`, `QueryParams`) which are merged together to create the model.

Also working on a unit test suite now.

sfentress:
This is tastable at http://authoring.concord.org/activities/9366

PR #53, opened 2019-02-07
Add mystery labels

sfentress:
This adds simple props for using or not using the mystery labels for organelles and substances. All view requests for the strings now go through a method in the organismsSpace model.

Adds a quick test, which would be uninteresting except that it exposes an odd MST behavior that we can't create more than one `OrganismsSpaceModel`, as we do in a test's `beforeEach` method, without MobX complaining that we are trying to create two of the same object, specifically not the `OrganismsSpaceModel` but the first `OrganismsRowModel` of each.

That is, the following code
 
    const model1 = OrganismsSpaceModel.create({});
    const model2 = OrganismsSpaceModel.create({});   // even with some other properties

throws

    [mobx-state-tree] Cannot add an object to a state tree if it is already part of the
    same another state tree. Tried to assign an object to '/rows/0', but it lives already
    at '/rows/0'

This could only be resolved by explicitly creating `OrganismsSpaceModels` with no rows.

Maybe @ekosmin has thoughts, but you don't need to spend time looking at it, it's just worth noting if it crops up again.

### DIFF
--- a/src/components/spaces/organisms/organelle-wrapper.sass
+++ b/src/components/spaces/organisms/organelle-wrapper.sass
@@ -1,3 +1,5 @@
+@import "../../vars"
+
 .model-wrapper
   position: relative
   cursor: default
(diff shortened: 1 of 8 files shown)


src/components/spaces/organisms/organelle-wrapper.sass
```diff
@@ -11,13 +13,16 @@
...
     top: 45px
-    right: 5px
+    right: 0
     min-width: 94px
-    height: 18px
-    background: #FFFFFFAA
+    height: 23px
+    color: #fff
+    background: $color-salmon-orange-dark2
     border: 1px solid #444
```
mklewandowski: This will need to be one of the new $color-simdata-carrot colors.  I removed all of the salmons as they were no longer part of the spec.

sfentress: :+1:

PR #56, opened 2019-02-08
162320674 show all data button graph

mklewandowski:
Add a button to the pop sim graph that allows the user to toggle between all data points and recent data points.

### DIFF
--- a/src/components/two-up-display.tsx
+++ b/src/components/two-up-display.tsx
@@ -4,6 +4,7 @@ import { BaseComponent, IBaseProps } from "./base";
 
 import "./two-up-display.sass";
 import { RightPanelType } from "../models/ui";
+import { ToolbarButton } from "../models/spaces/populations/populations";
 
 const titles: {[key in RightPanelType]: string} = {
   instructions: "Instructions",
@@ -22,7 +23,7 @@ interface IProps extends IBaseProps {
   rightPanel: React.ReactNode;
   spaceClass: string;
   rowNumber: number | undefined;
-  rightPanelButtons?: any;
+  rightPanelButtons?: ToolbarButton[];
 }
 interface IState {}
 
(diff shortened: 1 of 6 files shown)


src/components/two-up-display.tsx
```diff
@@ -90,11 +92,41 @@ export class TwoUpDisplayComponent extends BaseComponent<IProps, IState> {
...
           {this.props.rightPanel}
         </div>
-        {/* <div className={"footer " + this.props.spaceClass} data-test="right-footer"/> // removed for time being*/}
+        {this.renderRightPanelFooter()}
       </div>
     );
   }
 
+  private renderRightPanelFooter(){
+    if (this.props.spaceClass === "populations" &&
```
mklewandowski: Not sure if the button is going to eventually live in the footer, but we will need this space for other graph buttons eventually.  This button might end up floating on top of the graph in the upper right corner.

src/components/two-up-display.tsx
```diff
@@ -90,11 +92,41 @@ export class TwoUpDisplayComponent extends BaseComponent<IProps, IState> {
...
+
+      return (
+        <div className={"footer " + this.props.spaceClass} data-test="right-footer">
+          { buttons }
+        </div>
+      );
+    } else {
+      return null;
+    }
+
```
mklewandowski: Not sure if the button is going to eventually live in the footer, but we will need this space for other graph buttons eventually.  This button might end up floating on top of the graph in the upper right corner.

src/models/spaces/populations/mouse-model/mouse-populations-model.ts
```diff
@@ -195,6 +219,19 @@ export const MousePopulationsModel = types
...
+          const buttons = [];
+
+          buttons.push({
+            title: "Scale",
+            value: self.showMaxPoints,
+            action: (val: boolean) => {
+              self.toggleShowMaxPoints();
+            }
+          });
+
```
mklewandowski: I liked this idea of storing the toolbar buttons in an array model so I recycled it here.

sfentress: :+1: This worked well when I copied the pattern for the other buttons.

mklewandowski:
I don't have a strong opinion.  My gut reaction was to put everything in the footer for now and then restructure things if and when we have final styling and a prioritized story telling us to take care of it.  What do you think?

mklewandowski:
Looking at the v.2.3 UI spec now.  There is still nothing specified other than the styling of buttons in the footer (which is the styling I used for this button).  I think we should just put everything there for now.

sfentress:
👍

PR #57, opened 2019-02-08
Hawks soaring

sfentress:
This adds the circling hawks, and hawks now smoothly go down when eating, and back high when circling. This should remove some of the jitteriness that was caused by sudden size changes (though does not remove some jitteriness caused by turning suddenly to follow prey--if folks don't like the way the hawks turn, we can add some smoothing there too.)

### DIFF
--- a/src/models/spaces/populations/mouse-model/hawks.ts
+++ b/src/models/spaces/populations/mouse-model/hawks.ts
@@ -2,6 +2,8 @@ import { BasicAnimal, Species, Trait, Agent } from "populations.js";
 
 class Hawk extends BasicAnimal {
 
+  private _circleCount = 0;
+
   constructor(args: any) {
     super(args);
     this.label = "Hawk";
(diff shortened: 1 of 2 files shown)


src/models/spaces/populations/mouse-model/hawks.ts
```diff
@@ -39,15 +89,103 @@ export const hawkSpecies = new Species({
...
+      useIf: (agent: Agent) => agent.get("height") <= 8
+    }, {
+      image: {
+        path: "assets/curriculum/mouse/populations/hawk.png",
+        scale: 0.225,
+        anchor: {
+          x: 0.5,
+          y: 0.2
+        }
+      },
```
mklewandowski: I like the swoop.  It still feels a little subtle (don't think I would notice it unless I looked for it).  I probably would have incremented by .01 instead of .005 to make it more pronounced, but probably not worth investigating/adjusting now.

src/models/spaces/populations/mouse-model/mice.ts
```diff
@@ -393,7 +393,7 @@ export function getMouseSpecies(model: MousePopulationsModelType) {
           }, {
             image: {
               path: "assets/curriculum/mouse/populations/sandrat-dark.png",
-              scale: 0.2,
+              scale: 0.16,
               anchor: {
```
mklewandowski: In general, I think these new scales are spot on.  I find the sim to be much more coherent and readable with these smaller scales.  👍

src/models/spaces/populations/mouse-model/hawks.ts
```diff
@@ -1,16 +1,66 @@
...
+        y: loc.y + dy
+      })) {
+        newDir = this.get("direction") + (delta / 15);
+        this.set("direction", newDir);
+      }
+    }
+    this._circleCount += 0.01;
+    return this.move(speed);
+  }
+
```
mklewandowski: This movement handling all seems good.  Obviously we still have the jitters due to eating, but there is only so much we can do without radically changing what happens when a hawk eats a mouse (a particle effect, pop-up text, or some small animation effect to convey that a mouse has been consumed would be cool, but definitely not on the table for this round).  I actually like when the hawk movement brings the hawks slightly outside the bounds of the sim.  I think this looks neat.

sfentress: Yes, I actually would have liked to explore having the bounds of the model being larger than the bounds of the view, but that's tricky to do, since people wanted students to be able to see every mouse, to avoid the case where a mouse of a different color was outside the field of view and causing confusion with graphs or breeding.

The way to solve that would be to have the bigger model, and to have barriers to stop the mice from exiting the visible view. But right now we can't have different barriers for different species, so the hawks would be bound by it too.

sfentress:
I just realized after merging that populations.js *does* have a `getSize` function that could be defined per agent, so I could have used that instead of defining 10 different image scaling rules... Not going to worry about this now, but maybe if we update the hawk image with a sprite, because it would be a bore to define 10 different sprites...

Speaking of which, I realized this because I started adding [issues to the populations.js repo](https://github.com/concord-consortium/populations.js/issues), to remind us of things we want to add to the library. I was going to add "Make a way for an agent to define its size" and then I went to check if maybe it already existed... Point is, though, if you hit pain points with populations.js, add them to the issues tracker.

mklewandowski:
Ah, very cool.  Yeah, I think there is eventually a wing flapping hawk coming (I remember seeing a 2 frame animation from the artist).  Would be a good time to re-investigate the hawk handling.

PR #62, opened 2019-02-12
Insure we never get wrong mouse color population, reduce population when in wrong environment

sfentress:
Previously we were seeing occasional "wrong" mice populations, e.g. all brown mice with hawks in white environment.

This seemed to be being caused by:

1. Limited vertical movement (this has since been fixed)
2. Bugs in code that was supposed to reduce "wrong" color's mating chance.

The mating chance code has been fixed, but it seemed too powerful to run all the time. Now it just runs if the user eliminates one population and then switches environments.

With the fixed vertical movement, we generally don't get wrong results, but this also adds a bit of insurance to make sure we never eliminate the wrong population.

[#163911808]

### DIFF
--- a/src/models/spaces/populations/mouse-model/hawks-mice-interactive.ts
+++ b/src/models/spaces/populations/mouse-model/hawks-mice-interactive.ts
@@ -275,14 +275,32 @@ export function createInteractive(model: MousePopulationsModelType) {
       }
     }
 
+    // Insurance: Ensure there are always at least two rabbits of the correct color.
+    // We need to catch it before there are zero. Once there are zero, we can't create any out
+    // of thin air. (E.g. user eliminated one population then changed environment)
+    if (environmentColor === "white" && numWhite > 1 && numWhite < 3) {
+      for (let i = 0; i < 2; i++) {
+        addAgent(mouseSpecies, [], [createColorTraitByGenotype("a:C,b:C")]);
+      }
+    } else if (environmentColor === "neutral" && numTan > 1 && numTan < 3) {
+      for (let i = 0; i < 2; i++) {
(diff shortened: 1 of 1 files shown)


src/models/spaces/populations/mouse-model/hawks-mice-interactive.ts
```diff
@@ -275,14 +275,32 @@ export function createInteractive(model: MousePopulationsModelType) {
       }
     }
 
+    // Insurance: Ensure there are always at least two rabbits of the correct color.
```
kswenson: "rabbits"? 😉

src/models/spaces/populations/mouse-model/hawks-mice-interactive.ts
```diff
@@ -275,14 +275,32 @@ export function createInteractive(model: MousePopulationsModelType) {
...
+    // We need to catch it before there are zero. Once there are zero, we can't create any out
+    // of thin air. (E.g. user eliminated one population then changed environment)
+    if (environmentColor === "white" && numWhite > 1 && numWhite < 3) {
```
kswenson: Is there a reason to prefer this to `numWhite === 2`?

sfentress: Good point! It was a window that I had been changing, and hadn't even really twigged that it had become === 2...

src/models/spaces/populations/mouse-model/hawks-mice-interactive.ts
sfentress: Whatever. They're all still rabbits to me. Especially since they hop now. 🐰 😝

PR #63, opened 2019-02-12
162622045 graph button styling

mklewandowski:
Add styling to the show all/show recent graph data button.  This required some refactoring as this button needed to be shown in the upper-right corner, not in the toolbar.  I still like the idea of having an array of `rightPanelButtons` that we build on a per space basis.  However, some extra tweaking was required to specify how to place a button like the show all/show recent button that gets placed somewhere outside of the toolbar.  For now, I was thinking of these as floating buttons - buttons that float somewhere in the panel which would be determined by a CSS class tied to each button.  Obviously lots of ways of doing this, and I'm not eager to make an unusual case to handle a single button, but I was assuming something like this will probably come up again.

### DIFF
--- a/src/components/two-up-display.tsx
+++ b/src/components/two-up-display.tsx
@@ -93,23 +93,49 @@ export class TwoUpDisplayComponent extends BaseComponent<IProps, IState> {
           {this.props.rightPanel}
         </div>
         {this.renderRightPanelFooter()}
+        {this.renderFloatButtons()}
       </div>
     );
   }
 
-  private renderRightPanelFooter(){
-    if (this.props.spaceClass === "populations" &&
-        this.props.selectedRightPanel === "data" &&
-        this.props.rightPanelButtons) {
-      const buttons = this.props.rightPanelButtons.map( (button: any) => {
+  private renderRightPanelFooter() {
+    if (this.props.rightPanelButtons) {
+      let buttons = null;
+      buttons = this.props.rightPanelButtons.map( (button: any) => {
         const type = button.type || "button";
-        let title = button.title;
-        if (button.title === "Scale") {
-          title = button.value ? "Show Recent Data" : "Show All Data";
-        }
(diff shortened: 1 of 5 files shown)


src/components/two-up-display.tsx
```diff
@@ -93,38 +93,63 @@ export class TwoUpDisplayComponent extends BaseComponent<IProps, IState> {
           {this.props.rightPanel}
         </div>
         {this.renderRightPanelFooter()}
+        {this.renderFloatButtons()}
       </div>
```
mklewandowski: Float buttons are get absolute positioning and float somewhere in the right panel.

src/models/spaces/populations/mouse-model/mouse-populations-model.ts
```diff
@@ -226,10 +226,13 @@ export const MousePopulationsModel = types
 
           buttons.push({
             title: "Scale",
+            type: "float-button",
```
mklewandowski: Marking a button as type `float-button` means we will handle it differently than the other buttons and try to position it somewhere floating in the right panel.

src/models/spaces/populations/mouse-model/mouse-populations-model.ts
```diff
@@ -226,10 +226,13 @@ export const MousePopulationsModel = types
...
             value: self.showMaxPoints,
             action: (val: boolean) => {
               self.toggleShowMaxPoints();
-            }
+            },
+            floatCorner: "upper-right",
+            section: "data"
           });
```
mklewandowski: Use `floatCorner` to specify where we will position the button (for now I'm thinking of the buttons as floating in one of the 4 corners, but this could be made more generic for things like the top-center, bottom-center, etc.)  `section` specifies which of the 3 right panel sections shows the buttons.  I'm assuming each section might have it's own set of toolbar buttons

PR #65, opened 2019-02-13
Add graph axes labels and scale for weeks

sfentress:
This adds labels to the chart axes, following the pattern in Dat Cross Dam, and also scales the x axes so that we are showing weeks, where each step is 10 weeks.

This required a small addition to the dataSet model to add an initial max A1 value.

### DIFF
--- a/src/models/spaces/charts/chart-data-set.ts
+++ b/src/models/spaces/charts/chart-data-set.ts
@@ -122,14 +122,20 @@ export const ChartDataSetModel = types
       if (self.fixedMaxA1 !== undefined && self.dataPoints.length <= self.fixedMaxA1) {
         return self.fixedMaxA1;
       } else if (!self.visibleDataPoints || self.visibleDataPoints.length === 0) {
-        if (self.maxPoints) {
+        if (self.initialMaxA1){
+          return self.initialMaxA1;
+        } else if (self.maxPoints) {
           return self.maxPoints;
         } else {
           return defaultMax;
         }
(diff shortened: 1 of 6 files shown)


src/models/spaces/charts/chart-data-set.ts
```diff
@@ -73,12 +73,17 @@ export const ChartDataSetModel = types
...
+    // initial maximum which is not simply the value of maxPoints, which is otherwise the default.
+    initialMaxA1: types.maybe(types.number),
     // expandOnly is used for y-axis scaling. When requesting min/max point values,
     // if this is set the a2 / y axis max returns the max of the full data set, not just the visiblePoints
     expandOnly: false,
     fixedLabelRotation: types.maybe(types.number),
     dataStartIdx: types.maybe(types.number),
-    stack: types.maybe(types.string)
+    stack: types.maybe(types.string),
+    a1AxisLabel: types.maybe(types.string),
```
dstrawberrygirl: The pattern used earlier is the field description followed by the axis, (thingA1, thingA2), so fixedMaxA1, fixedMinA2, etc. I realize now that the labels I added in the dat-cross-dam repository broke this naming rule, (by going with a1Thing), so this one's my fault. I think it would make sense to fix this to be axisLabelA1, axisLabelA2, to follow the same pattern as the other property names, if that makes sense to you.

sfentress: :+1: updated now

PR #69, opened 2019-02-19
163461032 relocate mouse indicator

mklewandowski:
Adjust position of the `CollectButton` React component when it is being used as the selected mouse indicator at the organism level to match UI spec.  This required changing the creation point, so we could position the `CollectButton` on top of other elements in the z-order (previously we were creating it in the `OrganismsContainer`, now we are doing so in organism space) .  I also adjusted the collect button styling (both in the left nav panel and in the organism view) so that it has styling that is up to date with the latest spec (sizing, positions, hover states, etc. had drifted from the spec updates).


mklewandowski:
Agree.  I'm not crazy about the organizational location of the `CollectButton`, but to visually move it up on top of everything required in the spec, we would need to change the positioning style of everything in the two-up and 4-up displays which felt like a chaotic set of changes.  I wish the mouse indicator was just part of the bottom toolbar, but we're so limited on space that that would never be an option.

PR #70, opened 2019-02-19
163965874 style organism controls

mklewandowski:
Update the organism toolbar and buttons to match the UI spec.  Add organism level cursors for the different modes (assay, inspect, add) that the user can access.


mklewandowski:
Right.  Loading all of these SVGs into the index.html file is a remnant of a CLUE design pattern and maybe not ideal once we expand the app to include more units.  I agree we should think about a cleaner method of organizing these assets.

PR #72, opened 2019-02-21
Author graph buttons

sfentress:
This adds authoring to enable/disable the graph buttons.

https://connected-bio-spaces.concord.org/branch/author-graph-buttons/?authoring

### DIFF
--- a/src/authoring-schema.json
+++ b/src/authoring-schema.json
@@ -192,6 +192,18 @@
               }
             }
           }
+        },
+        "enableColorChart": {
+          "title": "Enable color chart",
+          "type": "boolean"
+        },
+        "enableGenotypeChart": {
+          "title": "Enable genotype chart",
+          "type": "boolean"
+        },
+        "enableAllelesChart": {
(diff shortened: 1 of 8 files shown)


sfentress:
That's fair that you thought that they should be there. My mental model, instead, was all the model options were together, and the graph changes came after that (still in the Populations section, of course).

The main issue is that, afaik, we can't have subheadings without nesting the json schema, and we can't create a flat model with a nested json structure. Since I really *wanted* to have subheadings for some parts of the authoring (e.g. "initial mouse population") I added a hacky flattening of the json, which is why all the `MousePopulationsModel` props have a dot notation. But I didn't want to keep pushing that, since the dot notation is ugly and I wish it didn't have it, and it seems silly that we have it just for a stylistic concern for the authoring page.

The better solution would be to be able to have arbitrary subheadings that aren't reflected in the json, but I don't know if there's an easy way to do that with what we have, so it may take jumping into a more in-depth re-working of the authoring page.

PR #77, opened 2019-02-26
Add graph styling to organism view

mklewandowski:
Add styling to the organism level graph based on the updated UI spec.


sfentress:
:+1: Looks good.

There's a potential question about the ordering of the labels, but this is good to pull in now.

mklewandowski:
Agreed.  I tried to match the spec for the label order, but this merits discussion.

PR #79, opened 2019-02-27
Click to add mice to the organism panel

sfentress:
This updates the UX from requiring the user to click twice (once on the backpack and once in the row) to single click on the backpack, which auto-populates the first available row, if any.

This also deselects mice on space changes, and on row clearing, to prevent the user from needing to click to deselect a mouse before they can add it.

There is currently no purpose now to a backpack mouse being in a selected state. We can discuss at the meeting whether we should remove this behavior entirely, or if it will come up in a later interaction.

### DIFF
--- a/src/components/collect-button.tsx
+++ b/src/components/collect-button.tsx
@@ -47,7 +47,7 @@ export class CollectButtonComponent extends BaseComponent<IProps, IState> {
           </svg>
         </div>
         <div className="collect-button-outline">
-          <div className={buttonClass} onClick={handleMouse} data-test="stored-mouse-class">
+          <div className={buttonClass} onMouseDown={handleMouse} data-test="stored-mouse-class">
             <div className={innerOutlineClass} data-test="inner-outline">
               <img src={mouse.baseImage} className="icon" data-test="stored-mouse-image"/>
               <div className="label" data-test="stored-mouse-label">{mouse.label}</div>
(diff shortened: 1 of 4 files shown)


src/components/collect-button.tsx
```diff
@@ -47,7 +47,7 @@ export class CollectButtonComponent extends BaseComponent<IProps, IState> {
           </svg>
         </div>
         <div className="collect-button-outline">
-          <div className={buttonClass} onClick={handleMouse} data-test="stored-mouse-class">
+          <div className={buttonClass} onMouseDown={handleMouse} data-test="stored-mouse-class">
             <div className={innerOutlineClass} data-test="inner-outline">
```
mklewandowski: onMouseUp would make the mouse action feel more similar to our other buttons and clickable objects where the action doesn't occur until the mouse up event fires.

mklewandowski:
Also, removing or commenting out lines 39 and 40 in `collect-button.tsx` will prevent clicks on mice in the backpack when both organism rows are full from showing the selected/deselected states.  More specifically, these two lines:
    buttonClass += this.isDeselected() ? " deselected" : "";
    buttonClass += this.isSelected() ? " selected" : "";

Obviously, we would still need to decide if we want to rip out the CSS classes in the Sass file, but it at least prevents strange styling on the buttons after we've filled both organism level rows with a mouse.

sfentress:
Cool. I'm not sure if you see the comments I make in the commits, but the onMouseDown was an explicit decision for the users who expect to be able to drag mice into the slots, as this way it happens the moment they try to drag, instead of not happening at all. Also, yes, leaving the selected states in was so that we could discuss it at the meeting.

mklewandowski:
I didn't see that commit comment.  Makes sense.

PR #80, opened 2019-03-01
Amino Acid Info Box tweaks

sfentress:
Add a checkbox for showing the info box, make the amino-acid chain visible by default, small CSS tweaks.

### DIFF
--- a/src/components/spaces/proteins/protein-viewer.sass
+++ b/src/components/spaces/proteins/protein-viewer.sass
@@ -1,23 +1,47 @@
 .protein-viewer
   display: flex
-  align-items: center
   position: initial !important
 
 .proteins
   display: flex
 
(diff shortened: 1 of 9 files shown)


src/components/spaces/proteins/protein-viewer.sass
```diff
@@ -1,23 +1,47 @@
...
   bottom: 5px
 
-.options
-  position: absolute
-  bottom: 5px
-  display: flex
-  flex-direction: column
-  align-items: flex-start
+.info-and-options
+  font-size: 0.8em;
```
mklewandowski: I know we bounce between em and px - at some point we should try to unify these.  Not urgent, just hoping that typing this will make it stick in my head for future reference.

PR #81, opened 2019-03-04
Load authoring data into form

sfentress:
This is a fairly simple change which takes the authored data from the url, plus the authoring defaults, and passes them to the authoring form to be used as the initial form state.

It also now allows `?authoring` to be in a url with the other parameters, instead of requiring it to be the only url param, to allow it to be added to existing authoring urls. This allows us to take the authored url

http://localhost:8080/?%7B"curriculum"%3A"mouse"....

and get the authoring page version simply by inserting "authoring" after
the ?

http://localhost:8080/?authoring%7B"curriculum"%3A"mouse"

This is a little weird, because unescaped it looks like `?authoring{"curriculum": "mouse", ...}` but it seemed the best way both to be backwards-compatible with just using `?authoring` by itself, and also making it simple to see the authoring page of a model.

### DIFF
--- a/src/utilities/url-params.ts
+++ b/src/utilities/url-params.ts
@@ -7,13 +7,15 @@ export interface QueryParams extends ConnectedBioAuthoring {
 let params: QueryParams;
 
 try {
-  const queryString = location.search.length > 1 ? decodeURIComponent(location.search.substring(1)) : "{}";
+  const queryString = location.search.length > 1
+    ? decodeURIComponent(location.search.substring(1).replace("authoring", ""))
+    : "{}";
   params = JSON.parse(queryString);
 } catch (e) {
   params = {};
 }
 
-if (location.search === "?authoring") {
+if (location.search.indexOf("authoring") > -1) {
   params.authoring = true;
 }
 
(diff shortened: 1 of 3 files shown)


src/utilities/url-params.ts
```diff
@@ -7,13 +7,15 @@ export interface QueryParams extends ConnectedBioAuthoring {
...
+  const queryString = location.search.length > 1
+    ? decodeURIComponent(location.search.substring(1).replace("authoring", ""))
+    : "{}";
   params = JSON.parse(queryString);
 } catch (e) {
   params = {};
 }
 
-if (location.search === "?authoring") {
+if (location.search.indexOf("authoring") > -1) {
```
mklewandowski: I don't expect this to happen, but if the string "authoring" appears in the instructions, then we'll reroute back to the authoring page when the user hits submit.

sfentress: Yeah, I know. I went back and forth between being worried about that, even though the string doesn't currently appear in the authoring spec, and thinking that folks might put it after the `%` in `?%7B"curriculum"%3A"mouse"....`, and wanting to still support them if they did that.

But I didn't think about the remote possibility that someone might author instructions with the word "authoring" in it. It's a slim chance, but a good point. I'll fix it so that `authoring` must come right at the beginning.

PR #82, opened 2019-03-06
Udate Remove Hawks button state when user resets simulation

mklewandowski:
We were not updating the state when the user reset the pop sim and the "add hawks" button would display as "remove hawks" to the user.

[#164438603]


sfentress:
:+1:

PR #83, opened 2019-03-12
Use integer values on Pop Graph to avoid graph resizing

mklewandowski:
The population graph was resizing (wiggling) as it switched between x axis labels with different string lengths.  This was caused by some axis label values being integers and others having decimals (such as 180 and 180.5).  Changes round all axis label values on the line graph to eliminate the graph resizing and wiggle effect.


sfentress:
Looks good

PR #84, opened 2019-04-05
Fix populations graph button selection states

ekosmin:
The chart type was not an observable value, so updating it didn't cause a re-render.

I read this to figure out how to fix it: https://github.com/mobxjs/mobx-state-tree#volatile-state

### DIFF
--- a/src/models/spaces/populations/mouse-model/mouse-populations-model.ts
+++ b/src/models/spaces/populations/mouse-model/mouse-populations-model.ts
@@ -5,6 +5,7 @@ import { ToolbarButton } from "../populations";
 import { ChartDataModel } from "../../charts/chart-data";
 import { hawkSpecies } from "./hawks";
 import { ChartAnnotationModel, ChartAnnotationType } from "../../charts/chart-annotation";
+import { observable } from "mobx";
 
 const dataColors = {
   white: {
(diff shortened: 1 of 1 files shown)


ekosmin:
It's kind of interesting! It was when `chartType` got pulled into volatile state:

https://github.com/concord-consortium/connected-bio-spaces/commit/fb897824b4bcdf6b3c435087ba0295041121ccf9#diff-1e6687d24a05cb82f7c73062451a060eR199

But it wasn't marked observable, so it wasn't reactive. That meant that this `value` wasn't recalculated when the `chartType` changed:

https://github.com/concord-consortium/connected-bio-spaces/commit/fb897824b4bcdf6b3c435087ba0295041121ccf9#diff-1e6687d24a05cb82f7c73062451a060eR470

Covered in more detail here https://github.com/mobxjs/mobx-state-tree#volatile-state: 

And here's an example of how to do your own volatile state not using an observable (but if you do this make sure the local state will never be used in a computed value first and bear in mind it won't be reactive!):

```
// if not using an observable then make sure your local state is NOT part of a view getter or computed value of any kind!
// also changes to it WON'T be reactive
const Todo = types.model({}).extend(self => {
    let localState = 3
```

This was a problem because `chartType` *is* used in a computed function:

https://github.com/concord-consortium/connected-bio-spaces/commit/fb897824b4bcdf6b3c435087ba0295041121ccf9#diff-1e6687d24a05cb82f7c73062451a060eR325

kswenson:
It seems like there ought to be an MST-only solution that doesn't require the use of MobX's `observable.box`. As I understand it the subtlety is that `chartType` is both initialized from model values but also settable by the user. I'm curious whether something like this would work:
```typescript
  .volatile(self => ({
    userChartType: undefined as (ChartType | undefined)
  }))
  .views(self => ({
    get chartType() {
      return self.userChartType ? self.userChartType :
            self.enableColorChart ? "color" :
            self.enableGenotypeChart ? "genotype" :
            self.enableAllelesChart ? "alleles" : "color";
    }
  }))
  .actions(self => ({
    setChartType(chartType: ChartType) {
      self.userChartType = chartType;
    }
  }))
```

PR #85, opened 2019-04-06
Removes MobX-specific features from chart type setting

ekosmin:
Following up on @kswenson's suggestion to remove MobX-specific features from `chartType` functionality. I agree that this is cleaner than setting up observers in volatile state.

http://connected-bio-spaces.concord.org/branch/mst-only-charts/

### DIFF
--- a/src/models/spaces/populations/mouse-model/mouse-populations-model.ts
+++ b/src/models/spaces/populations/mouse-model/mouse-populations-model.ts
@@ -190,7 +190,8 @@ export const MousePopulationsModel = types
     "showMaxPoints": false,
     "enableColorChart": true,
     "enableGenotypeChart": true,
-    "enableAllelesChart": true
+    "enableAllelesChart": true,
+    "userChartType": types.union(types.undefined, ChartTypeEnum)
   })
   .volatile(self => ({
     hawksAdded: false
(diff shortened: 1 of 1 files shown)


src/models/spaces/populations/mouse-model/mouse-populations-model.ts
```diff
@@ -190,7 +190,8 @@ export const MousePopulationsModel = types
     "showMaxPoints": false,
     "enableColorChart": true,
     "enableGenotypeChart": true,
-    "enableAllelesChart": true
+    "enableAllelesChart": true,
+    "userChartType": types.union(types.undefined, ChartTypeEnum)
```
kswenson: Two thoughts:
1. Wouldn't `types.maybe(ChartTypeEnum)` be a simpler way of saying the same thing?
2. I suggested putting `userChartType` in the volatile section because it was in the volatile section before. Does putting it in the model have side effects on document storage in the same way it would in CLUE?

ekosmin: Shoot, good point about `types.many`.

As for volatile vs. model: I put it back in the model since that's where it was before Sam moved it out, which I assumed he did only to handle defaults more easily. But you're right, I'm not sure if there are side effects, so I'll make it volatile - especially because it seems like there wouldn't be any reason to store this info.

ekosmin: Fixed to match your original suggestion!

It is strange that you need to use `undefined as (ChartType | undefined)` instead of `Types.maybe(ChartTypeEnum)`, but I guess this is because volatile state doesn't support getters/setters.

ekosmin:
Shoot I forgot to merge this PR...I do believe I've resolved all your requests, does that seem correct @kswenson?

kswenson:
👍 Looks good

PR #86, opened 2019-05-06
Release 1.0.2

ekosmin:
@sfentress I'd like to release the population graph button changes from a long time ago into production. I noticed this repo hasn't been following all the release suggestions from the README, but I'd like to do it for this release if that sounds good to you. So if this is approved, I'll merge to master, then merge to production and create a tag.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,12 @@
 # Changelog
 
+## Version 1.0.2 - released May 6, 2019
+
+- Population graph button states update correctly
+
 ### Asset Sizes
 
-| File | Size | % Increase from Previous Release |
+| File | Size | % Change from Previous Release |
 |---|---|---|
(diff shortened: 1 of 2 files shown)


ekosmin:
Totally agree, thanks Sam! Right, all the production steps just came out of the starter repo and I agree it would be good to discuss them. Just some general thoughts:

1) Agreed that putting all the commits in is overkill, I usually just use that section to summarize release notes for PIs (I end up copying them into the tag release, which is maybe the place where they're more useful)
2) I like the size changes to make sure I haven't blown up the size of the library without meaning to, but there may be a better way of tracking that
3) Good point about the release number in the package...we're really tracking user-facing versions, not program-facing versions, and there's no reason to mix those together. It does help me keep track of the version a little better in my head

Feels like there is some room for improvement but I'll merge this for now

PR #92, opened 2019-09-05
switch travis to trusty and remove sudo config

scytacki:


scytacki:
What is weird is that the output from cypress has changed:
This is what it used to look like on a successful build:
https://travis-ci.com/concord-consortium/sizing-demo-interactive/jobs/231500407#L666-L671

And now it looks like:
https://travis-ci.com/concord-consortium/connected-bio-spaces/builds/126109548#L1113-L1117

There doesn't seem like there is any 'verifying cypress can run'.  Probably this doesn't matter, but  I don't understand why.

PR #108, opened 2020-01-21
Add nest pair display and mouse selection to breeding sim

mklewandowski:
Add mouse nest pairs and related toolbar buttons to breeding space.  Final art and styling will require update once assets and spec are finalized.


src/components/spaces/breeding/nest-pair.tsx
```diff
@@ -0,0 +1,196 @@
...
+        return "right-bottom";
+      default:
+        return "";
+    }
+  }
+
+  private getNestImage = () => {
+    switch (this.props.positionIndex) {
+      case 2:
+        return "../../../assets/curriculum/mouse/breeding/nesting/left-top.png";
```
sfentress: I think if you replace all `../../../assets/` in .ts/.tsx files with `assets/` then it should correctly find the relative path, and will work in branches. See other .tsx files that reference files in `assets/`.

Currently the deployed version at http://connected-bio-spaces.concord.org/branch/nesting-pairs/ doesn't show any of the new assets.

src/models/spaces/breeding/breeding.ts
```diff
@@ -17,7 +40,12 @@ export const BreedingModel = types
     motherGamete: types.maybe(types.string),
     fatherGamete: types.maybe(types.string),
     rightPanel: types.optional(RightPanelTypeEnum, "instructions"),
-    instructions: ""
+    instructions: "",
+    showSexStack: false,
+    showHeteroStack: false,
+    interactionMode: types.optional(BreedingInteractionModeEnum, "none"),
```
sfentress: This is something to confirm with Frieda, but I believe we had discussed entering the space with interaction mode already on "breed." Otherwise the user has nothing they can do when they enter, unless they click on a mode.

Separately (but related) I think we might have decided that the user must always be in one of three modes here, so "none" might be removed, and clicking on a selected button does nothing.

src/components/authoring.tsx
```diff
@@ -65,9 +65,16 @@ export const defaultAuthoring: ConnectedBioAuthoring = {
   },
   breeding: {
     instructions: "",
-    breedingType: "litter"
+    breedingType: "litter",
+    nestPairs: [
```
sfentress: Because of the fact that we're going to be randomizing the sex of the mice, I don't think that this needs to be a part of the authoring. I think the details of how to generate the pairs could be hard-coded in the `breeding.ts` model itself, perhaps in `createBreedingModel` (like `createOrganismsModel`).

(That is, unless you've understood that the authors might ever want to change the pairs, but so far I haven't heard this.)

src/components/spaces/breeding/nest-pair.tsx
mklewandowski: Ah thanks!  I was just looking at the deployed version and noticed this.

mklewandowski: Fixed.

src/models/spaces/breeding/breeding.ts
mklewandowski: I set the default mode to "breed".  I'd like to check with Michael and others about ensuring that a mode is always set (since the mock-up implied that there could be a "none" state).  Might very well be something that appears in the mock-ups once they are updated.

src/components/authoring.tsx
mklewandowski: I moved the nestPair creation to createBreedingModel.

PR #110, opened 2020-01-22
Randomize nest pair mouse sex

mklewandowski:
Randomize the sex of the mouse pairs when we create them in the breeding model.

### DIFF
--- a/src/models/spaces/breeding/breeding.ts
+++ b/src/models/spaces/breeding/breeding.ts
@@ -31,37 +31,23 @@ export type INestPair = Instance<typeof NestPair>;
 
 export function createBreedingModel(breedingProps: any) {
   if (!breedingProps.nestPairs || breedingProps.nestPairs.length === 0) {
-    const numPairs = 6;
-    const sexpairs = [];
-    for (let i = 0; i < numPairs; i++) {
-      Math.random() >= 0.5 ? sexpairs.push({ls: "male", rs: "female"}) : sexpairs.push({ls: "female", rs: "male"});
-    }
-    breedingProps.nestPairs = [
-      { leftMouse: {sex: sexpairs[0].ls, genotype: "RC"},
-        rightMouse: {sex: sexpairs[0].rs, genotype: "RR"},
-        label: "Pair 1"
-      },
-      { leftMouse: {sex: sexpairs[1].ls, genotype: "CC"},
-        rightMouse: {sex: sexpairs[1].rs, genotype: "RR"},
-        label: "Pair 2"
-      },
-      { leftMouse: {sex: sexpairs[2].ls, genotype: "CC"},
-        rightMouse: {sex: sexpairs[2].rs, genotype: "CC"},
-        label: "Pair 3"
-      },
-      { leftMouse: {sex: sexpairs[3].ls, genotype: "RR"},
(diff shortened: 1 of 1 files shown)


src/models/spaces/breeding/breeding.ts
```diff
@@ -30,14 +30,39 @@ const NestPair = types.model({
...
+      { leftMouse: {sex: sexpairs[4].ls, genotype: "RC"},
+        rightMouse: {sex: sexpairs[4].rs, genotype: "RC"},
+        label: "Pair 5"
+      },
+      { leftMouse: {sex: sexpairs[5].ls, genotype: "CC"},
+        rightMouse: {sex: sexpairs[5].rs, genotype: "RC"},
+        label: "Pair 6"
+      },
+    ];
+  }
```
sfentress: This is fine, and could go in as-is. I wonder, however, if this is clearer:

```
const breedingPairs = [
      ["RC", "RR"],
      ["CC", "RR"],
      ["CC", "CC"],
      ["RR", "RR"],
      ["RC", "RC"],
      ["CC", "RC"]
    ];

    breedingProps.nestPairs = breedingPairs.map((pair, i) => {
      const leftSex = Math.round(Math.random());
      const rightSex = 1 - leftSex;
      return {
        leftMouse: {sex: leftSex, genotype: pair[0]},
        rightMouse: {sex: rightSex, genotype: pair[1]},
        label: `Pair ${i + 1}`
      };
    });
```

To me this is clearer, because the substantive part is purely `breedingPairs`, and this clarifies what they are. This array could also be easily brought in via authoring if we ever wanted to author the colors, or randomized if we wanted to randomize the placement etc.

However, the final result is the same, so whichever is clearer to you.

(Note above is aircoded, so there may be syntax or logic errors.)

mklewandowski: Yeah, this is a nice approach.  I'll rough this in and push another change.  Thanks!

mklewandowski: Changes pushed.

src/models/spaces/breeding/breeding.ts
```diff
@@ -31,37 +31,23 @@ export type INestPair = Instance<typeof NestPair>;
...
+      ["RR", "RR"],
+      ["RC", "RC"],
+      ["CC", "RC"]
     ];
+    breedingProps.nestPairs = breedingPairs.map((pair, i) => {
+      const leftSex = Math.round(Math.random());
+      const rightSex = 1 - leftSex;
+      return {
+        leftMouse: {sex: leftSex >= 0.5 ? "female" : "male", genotype: pair[0]},
+        rightMouse: {sex: rightSex >= 0.5 ? "female" : "male", genotype: pair[1]},
```
sfentress: Oh, sorry, I forgot that sex is defined as a string and not a number. This could be even easier as 

```
const leftSex = Math.random() < 0.5 ? "female" : "male";
const rightSex = leftSex === "female" ? "male" : "female";
```
However this is just a nitpick. PR looks good, totally good to go in.

mklewandowski: Actually that might be safer for the case where we land right on .5.  I'm going to add that tweak to be safe.

PR #121, opened 2020-01-29
Add authoring of graph and gametes buttons

mklewandowski:
Add ability for author to turn on or off the breeding level graph types and the breeding level inspect gametes button.  Note that, at present, enabling/disabling inspect gametes has no effect as this button is always disabled (this feature will be used once we implement inspecting gametes).

### DIFF
--- a/src/components/authoring.tsx
+++ b/src/components/authoring.tsx
@@ -65,7 +65,10 @@ export const defaultAuthoring: ConnectedBioAuthoring = {
   },
   breeding: {
     instructions: "",
-    nestPairs: []
+    nestPairs: [],
+    enableColorChart: true,
+    enableGenotypeChart: true,
+    enableSexChart: true
   }
 };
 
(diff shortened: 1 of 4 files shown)


src/components/authoring.tsx
```diff
@@ -65,7 +65,10 @@ export const defaultAuthoring: ConnectedBioAuthoring = {
   },
   breeding: {
     instructions: "",
-    nestPairs: []
+    nestPairs: [],
```
sfentress: While you're here, I think it would be good to drop `nestPairs`. This is because if we ever *do* want to add authoring for nestPairs (e.g. the number or colors of pairs), we'll want to give this some reasonable default. But if authors have already authored models, this blank default will be saved to their authoring, and will overwrite any default we give,

mklewandowski: Good call.

PR #122, opened 2020-01-30
Litter scroll bar

sfentress:
Adds the custom slider to the litters


sfentress:
1. The scroll wheel behavior is very hard to get working, because by using the slider, and not the native scrolling mechanisms, we don't get scroll events on the dom element. In order to get scroll events on the element, we need to use native scrolling, but then that gives us a native scroll bar that is not in the spec.
   If you can think of a way to get this working, I'd be very happy.
2. I believe the handle is supposed to be at the top when we're looking at the top-most litter, just like a regular scroll-bar handle. See screenshot from https://hy7r7v.axshare.com/#id=8tunqt&p=connectedbio_-_specs__v2_7_
Screen Shot 2020-01-30 at 1 40 25 PM

sfentress:
Better screenshot.
Screen Shot 2020-01-30 at 1 44 45 PM

mklewandowski:
@sfentress punting on scroll wheel seems fine. Not sure if my comment was clear, but I agree, content should be at top when handle is at top.  However, this is not what I'm seeing on my machine:  
image

PR #123, opened 2020-01-30
Allow scroll wheel to scroll litter content

mklewandowski:
Allow the mouse scroll wheel to scroll litter content

### DIFF
--- a/src/components/spaces/breeding/breeding-view.tsx
+++ b/src/components/spaces/breeding/breeding-view.tsx
@@ -155,7 +155,7 @@ export class BreedingView extends BaseComponent<IProps, IState> {
     const numLitters = activeBreedingPair.litters.length;
     const sliderResolution = numLitters > 50 ? 1 : numLitters > 30 ? 2 : numLitters > 10 ? 10 : 100;
     const sliderMax = Math.max((numLitters - 1) * sliderResolution, 0);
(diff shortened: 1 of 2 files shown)


src/components/spaces/breeding/breeding-view.tsx
```diff
@@ -148,6 +148,20 @@ export class BreedingView extends BaseComponent<IProps, IState> {
     );
   }
 
+  private handleWheel = (e: any) => {
+    const { breeding } = this.stores;
+    const { litterSliderVal } = this.state;
+    const activeBreedingPair = breeding.activeBreedingPair!;
+    const numLitters = activeBreedingPair.litters.length;
+    const sliderResolution = numLitters > 50 ? 1 : numLitters > 30 ? 2 : numLitters > 10 ? 10 : 100;
+    const sliderMax = Math.max((numLitters - 1) * sliderResolution, 0);
```
sfentress: This calculation could potentially be extracted into `sliderMax` state set in `getDerivedStateFromProps` or just simply a getter, now that `sliderMax` is being calculated in two places, but we could hold off on that until we find we need to change the calculation.

mklewandowski:
@sfentress I'm seeing huge jumps when I mouse wheel scroll using your suggested changes using my PC/mouse combo.  It makes the mouse wheel scrolling pretty unusable on my machine.

PR #124, opened 2020-02-03
Save breeding litters

sfentress:
This saves student data for the breeding space.

If any nest pair has more than 20 litters, we will just save the metadata for those litters, instead of saving every offspring. The metadata contains the number of litters, and the number of offspring of each genotype. The actual composition of each litter will be random when the student opens up their data again, but those constraints will be satisfied.

Added a simple test of this metadata expansion.

Also added a small dev feature for testing data saving: appending the url parameter `?saveToLocalStore` will save the user state under the local storage key `cbioSaveState` every time the data changes, and reloading the page will load that data. To clear it, clear it manually from the Application tab in the console.


sfentress:
Note:

The litters sizes would not be huge under normal usage. If a student saves 500 offspring for each pair, the total data size is about 250kB, which presumably ought to be fine. The issue is that there is nothing preventing a particularly bored student from sitting and clicking "breed" for twenty minutes, generating tens of thousands of offspring. Using the condensed format, the size of these tens of thousands of offspring will be essentially zero.

(Note: once the offspring are expanded, there may still be memory issues. This does not seek to address that.)

src/utilities/data.ts
```diff
@@ -42,3 +42,13 @@ export function downsample(points: DataPointType[], maxSize: number) {
...
+
+/**
+ * Durstenfelt in-place shuffle
+ */
+export function shuffle(arr: any[]) {
+  for (let i = arr.length - 1; i > 0; i--) {
+    const j = Math.floor(Math.random() * (i + 1));
+    [arr[i], arr[j]] = [arr[j], arr[i]];
+  }
+}
```
mklewandowski: Oddly, I also needed an array shuffle in my PR.  I pulled in the lodash version of shuffle (since we already have lodash as a dependency).  Not sure if you have a preference, but the lodash path wouldn't require this function.

PR #125, opened 2020-02-04
170858466 show hide gametes

mklewandowski:
Add ability to show and hide the mother/father gametes for each nest pair in the breeding view.  A few notes:
- the gamete ordering is, as requested, shuffled underneath each parent.  `litterShuffledGametePositions`, an array in each `NestPair` holds the shuffled gamete indices for each litter.
- the egg/sperm icon SVGs were added as SVG files to the `assets` folder and are displayed as background images via CSS.  This deviates from the prior application pattern of integrating SVGs but was needed to facilitate the egg icon hover state which includes a multi-color change (rather than a single color change which we handle elsewhere in the app with a CSS fill).  This is the only hover effect included in this PR.
- helper functions were added to `genetics.ts` which are used to get the HTML (which includes <sup> tags to handle the superscripts) for the genotypes.  Seemed like a good place to put these functions.


sfentress:
...all that said, I just realized that I'm not sure why we need to keep the gamete positions. Was that part of the story? It seems that if they want them to be random, they could just be random.

mklewandowski:
> ...all that said, I just realized that I'm not sure why we need to keep the gamete positions. Was that part of the story? It seems that if they want them to be random, they could just be random.

I was assuming that they were random positions, but that they needed to persist.  So that if a student looked at a litter and made note of the gametes and then went and did something else (looked at other litters, etc.) when they returned, the positions would still be intact.  Did I overthink this requirement?  It would be much easier if the randomization just occurred in the React component.

PR #129, opened 2020-02-05
Update allele names with new nomenclature

mklewandowski:
Update allele names with new nomenclature


sfentress:
...actually, looking at the screenshots, maybe it's less legible. So perhaps we should keep the `<sup>` format in those cases.

mklewandowski:
It would certainly be nice to remove those functions.  I agree that it's a little less clunky looking.  However it is also quite a bit harder to read.  And using <sup> renders the text so it is in sync with how it looks in the UI spec.  Anyway, I'm going to merge this in and then circle back to it when time permits to compare the two methods.

PR #136, opened 2020-02-08
Add gamete hover effects in breeding view

mklewandowski:
Add gamete hover effects into the breeding view when show gametes option is turned on.  Notes:
- a new `ArrowPanel` component takes in an array of `ArrowInfo` objects which specify each hover arrow and then renders the full set of arrows
- additional highlight properties have been added to the React state in the `BreedingView`component to help manage what object a user is hovering over and the subsequent effects that this has on other parts of the UI 
- additional props were added to the `StackedOrganism` to support the new mouse hover effects

### DIFF
--- a/src/components/spaces/breeding/breeding-view.tsx
+++ b/src/components/spaces/breeding/breeding-view.tsx
@@ -4,6 +4,7 @@ import Slider from "rc-slider";
 import { BaseComponent, IBaseProps } from "../../base";
 import { StackedOrganism } from "../../stacked-organism";
 import { gameteHTMLLabel } from "../../../utilities/genetics";
+import { ArrowPanel, ArrowInfo } from "./arrow-panel";
 
 import "./breeding-view.sass";
 import "rc-slider/assets/index.css";
@@ -11,6 +12,8 @@ import "rc-slider/assets/index.css";
 interface IProps extends IBaseProps {}
 interface IState {
   litterSliderVal: number;
+  offspringHightlightIndex: number;
+  parentHightlightIndex: number;
 }
 
 @inject("stores")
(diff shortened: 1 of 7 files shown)


src/components/spaces/breeding/breeding-view.tsx
```diff
@@ -128,19 +138,36 @@ export class BreedingView extends BaseComponent<IProps, IState> {
...
+                                          ? this.handleOffspringHoverExit(j)
+                                          : undefined}
+                            key={"org-cont" + j}>
+                            <StackedOrganism
+                              key={"org" + j}
+                              organism={org}
+                              organismImages={[org.nestImage]}
+                              height={60}
+                              showSelection={false}
+                              showGameteSelectionOnHover={breeding.showGametes && (currentLitter === (litterNum - 1))}
```
sfentress: I find the combination of `showGameteSelectionOnHover` and `showGameteSelection` to be a little confusing, because the hover state is something this container view already knows about, because if we're hovering we must have this org equal to `this.state.offspringHightlightIndex`.

I feel like `showGameteSelectionOnHover` is redundant. It adds a class that has an associated `:hover` state in the css, but that rule is always trumped by the addition of the `show` class which is passed down because of `showGameteSelection`, and which is always there if we are hovering.

I might be just confused, though.

src/components/spaces/breeding/breeding-view.tsx
```diff
@@ -201,6 +243,71 @@ export class BreedingView extends BaseComponent<IProps, IState> {
...
+    });
+    gameteRightPositions.forEach((position, i) => {
+      const startY = i % 2 === 1 ? gameteYStagger : 0;
+      const endY = baseEndY + (position % 2 === 1 ? yStagger : 0);
+      const startX = rightGameteXStart + gameteXOffset + i * gameteDelta;
+      const endX = mouseXStart + mouseXOffset + position * mouseXIncrement + arrowGap;
+      const visible = offspringHightlightIndex === position;
+      const arrow: ArrowInfo = { startX, endX, startY, endY, headRotation: 10, visible };
+      hoverArrows.push(arrow);
+    });
```
sfentress: This all currently works very well. I worry a little bit about how brittle it is. Did you consider getting the element position directly from the DOM?

It might have been possible to simply attach a `ref` to the elements in the gametes and the current litter, and then use `.getBoundingClientRect()` to get the exact position of the arrow start and ends relative to the viewport. Then this ought to just work even if we move the offspring images around.

But again, this does work, so maybe there's no need to change it, unless perhaps we find ourselves going back to make tweaks in the future.

src/components/spaces/breeding/breeding-view.tsx
mklewandowski: Good call.  My goal was to allow a caller to set the green gamete circle to be hoverable or to be shown due to some other interaction (which we need since a user can, for example, hover over a gamete icon which needs to turn the green gamete circle on) - hence the two values passed via props.  However, you're right, at present we have everything in the container state to effectively remove `showGameteSelectionOnHover` from the props and handle all of the green gamete circle states simply with `showGameteSelection`.

Anyway, I pushed a change to remove `showGameteSelectionOnHover`

src/components/spaces/breeding/breeding-view.tsx
mklewandowski: Agreed.  It is brittle.  I originally planned to try to access the positions by grabbing elements out of the DOM, but it felt like more work than it was worth so I went this route of baking in a handful of constants and doing a little math.  Maybe I'm wrong, however.  It might not be that bad to use a React ref to then pull the exact of the components to get the start/end locations needed to build the arrows.  

At this point in the sprint, I would vote that we merge this in for now, and I'll try to circle back to it in the next 48 hours if time permits (or add a refactor story to PT).  What do you think of a plan like that?

sfentress: :+1:

PR #143, opened 2020-02-12
Add breed level collection of parent and offspring mice

mklewandowski:
Add breed level collection of parent/offspring mice in both the nest and breeding views.  When collected, the mouse is cloned as a newly created backpack mouse and added to the backpack.  A reference to the original collected mouse is stored in the backpack store to allow highlighting the mice at the breed level.  The previous implementation of the nest view collection has been removed and replaced with this new model (so we no longer store a reference to the backpack mouse with the NestPair).

### DIFF
--- a/src/models/backpack.ts
+++ b/src/models/backpack.ts
@@ -25,7 +25,11 @@ export const BackpackModel = types
 
       isSelected(mouse?: BackpackMouseType) {
         return self.activeMouse && self.activeMouse === mouse;
-      }
+      },
+
+      cloneExists(id: string | undefined): boolean {
+        return (id && self.collectedMice.find(bpMouse => bpMouse.originMouseRefId === id) ? true : false);
+      },
     };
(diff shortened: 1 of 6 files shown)


src/models/backpack.ts
```diff
@@ -25,7 +25,11 @@ export const BackpackModel = types
 
       isSelected(mouse?: BackpackMouseType) {
         return self.activeMouse && self.activeMouse === mouse;
-      }
+      },
+
+      cloneExists(id: string | undefined): boolean {
+        return (id && self.collectedMice.find(bpMouse => bpMouse.originMouseRefId === id) ? true : false);
```
sfentress: Could be simplified as `return id && self.collectedMice.some(bpMouse => bpMouse.originMouseRefId === id)`

mklewandowski: Fixed

PR #145, opened 2020-02-13
Save panel state

mklewandowski:
Save the right panel state in the breeding view so users return to the proper panel state when reloading their work.


mklewandowski:
@sfentress Perhaps I misinterpreted.  I'll check with Frieda.

mklewandowski:
@sfentress yup, you're right.  I'll update so we restore the tab if we have data.

PR #148, opened 2020-02-14
Restore data panel if user began breeding

mklewandowski:
Restore the data panel on the breeding level if we open the user save state and they have begun breeding.  I went with checking if one of the nest pairs had been visited via the `hasBeenVisited` property (rather than checking if a litter array contains mice) since checking `hasBeenVisited` will also handle the following cases:
-user has bred mice and hit the reset button
-user has selected one or more pairs that they want to mark for breeding but hasn't hit breed button yet
Both felt like valid reasons to restore the data panel.  However, I'm open to changing this if we put it in front of users and it is confusing.

### DIFF
--- a/src/models/spaces/breeding/breeding.ts
+++ b/src/models/spaces/breeding/breeding.ts
@@ -244,6 +244,9 @@ export function createBreedingModel(breedingProps: any) {
       };
     });
   }
+  if (breedingProps.nestPairs && breedingProps.nestPairs.some((pair: INestPair) => pair.hasBeenVisited)) {
+    breedingProps.rightPanel = "data";
+  }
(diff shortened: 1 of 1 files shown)


src/models/spaces/breeding/breeding.ts
```diff
@@ -244,6 +244,9 @@ export function createBreedingModel(breedingProps: any) {
       };
     });
   }
+  if (breedingProps.nestPairs && breedingProps.nestPairs.some((pair: INestPair) => pair.hasBeenVisited)) {
```
kswenson:
```
  if (breedingProps.nestPairs?.some((pair: INestPair) => pair.hasBeenVisited)) {
```

PR #153, opened 2020-04-13
172274981 graph popups

mklewandowski:
This PR adds the following improvements to the line chart in the population view:
- filter duplicate x values so that tooltip popup at t = 0 doesn't show additional, unexpected values (with present design, x represents time so we should not have multiple samples at a given x or time value).
- fix tooltip popup colors so that colors match the graph lines and legend
- prevent duplicate values in x axis labels by doing the following: set min axis value to 12 when in "show all data" mode and change rounding of min value to avoid min val and first tick having same value
- round values shown in tooltip popup to use max of 2 decimal places

### DIFF
--- a/src/components/charts/line-chart.tsx
+++ b/src/components/charts/line-chart.tsx
@@ -178,6 +178,17 @@ export class LineChart extends BaseComponent<ILineProps, ILineState> {
       },
       annotation: {
         annotations: chartData.formattedAnnotations
+      },
+      tooltips: {
+        enabled: true,
+        callbacks: {
+          labelColor(tooltipItem: any, chart: any) {
+            return {
+              borderColor: "rgb(255, 255, 255)",
+              backgroundColor: chart.config.data.datasets[tooltipItem.datasetIndex].borderColor
+            };
+          },
+        }
       }
     });
     const w = width ? width : 400;
--- a/src/components/charts/line-chart.tsx
+++ b/src/components/charts/line-chart.tsx
@@ -167,7 +167,10 @@ export class LineChart extends BaseComponent<ILineProps, ILineState> {
             minRotation: chartData.dataLabelRotation,
             maxRotation: chartData.dataLabelRotation,
             userCallback: (label: any) => {
-              return Math.ceil(label);
+              return (label === minMaxValues.minA1
+                      ? Math.floor(label)
+                      : Math.ceil(label)
+              );
             },
           },
           scaleLabel: {
--- a/src/components/charts/line-chart.tsx
+++ b/src/components/charts/line-chart.tsx
@@ -185,6 +185,9 @@ export class LineChart extends BaseComponent<ILineProps, ILineState> {
       tooltips: {
         enabled: true,
         callbacks: {
+          label(tooltipItems: any, data: any) {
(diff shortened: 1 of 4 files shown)


src/components/charts/line-chart.tsx
```diff
@@ -167,7 +167,10 @@ export class LineChart extends BaseComponent<ILineProps, ILineState> {
             minRotation: chartData.dataLabelRotation,
             maxRotation: chartData.dataLabelRotation,
             userCallback: (label: any) => {
-              return Math.ceil(label);
+              return (label === minMaxValues.minA1
+                      ? Math.floor(label)
+                      : Math.ceil(label)
+              );
```
mklewandowski, line 173: This rounds down the min value to prevent duplicate values on x axis
image

src/components/charts/line-chart.tsx
```diff
@@ -178,6 +181,20 @@ export class LineChart extends BaseComponent<ILineProps, ILineState> {
       },
       annotation: {
         annotations: chartData.formattedAnnotations
+      },
+      tooltips: {
+        enabled: true,
+        callbacks: {
+          label(tooltipItems: any, data: any) {
+            return `(${tooltipItems.xLabel}, ${Math.round(tooltipItems.yLabel * 100) / 100})`;
+          },
```
mklewandowski, line 190: I chose to do the rounding here to preserve the data as long as we can and only adjust the final displayed text in the tooltip popup.

src/models/spaces/charts/chart-data-set.ts
```diff
@@ -228,7 +228,9 @@ export const ChartDataSetModel = types
     }
 
     function addDataPoint(a1: number, a2: number, label: string) {
-      self.dataPoints.push({ a1, a2, label });
+      if (!self.dataPoints.some(d => d.a1 === a1)) {
+        self.dataPoints.push({ a1, a2, label });
+      }
```
mklewandowski, line 233: This duplicate value filtering might be too aggressive.  It could cause slow-down or it could prevent some future functionality where we would want duplicate values for a given a1 (at present these data points represent values over time so inserting multiple values at a given a1 or t does not make sense).  The primary reason this is here is so that we do not duplicate the values at t = 0 (since we have a default set of values in the model and we add another set when the user presses start).  We could also add in a check to see if t ===0  (a1===0 at this point in the code) or stop the code from adding the duplicate values elsewhere.

sfentress: I agree that this seems a little heavy-handed for this small bug, and also seems the wrong place for the fix, to me. I feel like it would be better just to fix the issue that is causing it to put the data at time zero twice.

I forget the exact reason that we put the first data-point in before the model starts, but I think that that was explicitly desired. So we just shouldn't put the data there a second time when the model starts.

I think if you remove the entire listener `Events.addEventListener(Environment.EVENTS.START ...` in mouse-populations-model.ts`, it solves the problem. With a brief look, I haven't found any cases where we needed the data to be added on START.

mklewandowski: Agreed.  Fixed.

PR #159, opened 2020-04-22
Respond to "touchstart" events to select mice

sfentress:


src/components/spaces/populations/populations-container.tsx
```diff
@@ -187,7 +187,7 @@ export class PopulationsComponent extends BaseComponent<IProps, IState> {
     const populations = this.stores.populations;
     if (populations) {
       if (evt.agents && evt.agents.mice) {
-        if (evt.type === "click" && populations.interactionMode === "select") {
+        if ((evt.type === "click" || evt.type === "touchstart") && populations.interactionMode === "select") {
```
kswenson, line 190: To state what may be obvious, these aren't completely analogous. On a mouse device the `click` event occurs at the end of a `mousedown` => `mouseup` => `click` sequence, whereas on a touch device it should be `touchstart` => `touchend` => `mousedown` => `mouseup` => `click`. The fact that the `click` apparently doesn't make it here on touch devices suggests that some intervening handler is calling `preventDefault()` on one of the earlier events.

PR #163, opened 2020-04-23
Improve population stabalization

sfentress:
This makes the population stabilization more subtle by

1. Introducing fewer new mice
2. Introducing new mice as babies near other mice of their color, to make it seem like a normal birth


sfentress:
Note, this is just from `Fix how agents are added to new environment` onwards, other commits will be rebased out once #162 is merged.

PR #167, opened 2020-04-30
Populations pie charts

sfentress:
Adds populations pie charts, and authoring for whether to show the button ~(defaults to false for now)~. Pie charts currently just show initial and last states, they do not connect to the scrubber.

Edit: Since we won't be deploying this until after the current school year, there is no need to default to false, so I switched that.


src/models/spaces/populations/mouse-model/mouse-populations-model.ts
```diff
@@ -8,17 +8,17 @@ import { BackpackMouse, BackpackMouseType } from "../../../backpack-mouse";
...
   },
   neutral: {
-    mice: "#db9e26",
+    mice: "#bf8f53",
     environment: "rgb(241,216,168)",
     environmentHighlight: "rgba(241,216,168, .7)"
   },
   brown: {
-    mice: "#795423",
+    mice: "#5d3515",
```
mklewandowski, line 21: Did we have these colors wrong?

src/models/spaces/populations/mouse-model/mouse-populations-model.ts
```diff
@@ -304,7 +307,7 @@ export const MousePopulationsModel = types
     function addSettingsAnnotation(label: string, labelXOffset: number) {
       const now = interactive ? interactive.environment.date : 0;
       const timeSinceLastAnnotation = now - lastSettingsAnnotationDate;
-      const labelYOffset = timeSinceLastAnnotation < 30 ? 55 : 30;
+      const labelYOffset = timeSinceLastAnnotation < 30 ? 170 : 145;
```
mklewandowski, line 310: We might want to add constants for these hard-coded values.  I'll certainly never remember what they mean :)

src/models/spaces/populations/mouse-model/mouse-populations-model.ts
```diff
@@ -506,17 +512,6 @@ export const MousePopulationsModel = types
...
-            title: "Scale",
-            type: "float-button",
-            value: self.showMaxPoints,
-            action: (val: boolean) => {
-              self.toggleShowMaxPoints();
-            },
-            floatCorner: "upper-right",
-            section: "data"
-          });
-
```
mklewandowski, line 519: Glad to see this removed!  I was never happy with how I implemented this as a `float-button`.

src/utilities/color-utils.ts
```diff
@@ -22,3 +22,10 @@ export function hexToRGB(hex: string) {
     b: parseInt(convertedHexValue[3], 16)
   } : null;
 }
+
+export function colorIsLight(hex: string) {
+  const rgb = hexToRGB(hex);
+  if (!rgb) return false;
+  const brightness = ((rgb.r * 299) + (rgb.g * 587) + (rgb.b * 114)) / 1000;
+  return brightness > 150;
+}
```
mklewandowski, line 31: I like this utility function. I probably should have used something like this for the hover color on the graph labels.

src/components/spaces/populations/populations-charts.tsx
```diff
@@ -0,0 +1,163 @@
...
+    let data = this.props.chartData.dataSets.filter(ds => ds.display).map(ds => {
+        const point = ds.dataPoints[(initialData || !ds.dataPoints.length) ? 0 : ds.dataPoints.length - 1];
+        const value = point && !isNaN(point.a2) ? point.a2 : 0;
+        return {
+          label: ds.name,
+          value,
+          color: ds.color as string
+        };
+    }) as PieChartData[];
+    data = data.filter(d => d.value);
```
mklewandowski, line 127: I wonder if the pie chart data needs to specify how many decimal digits to use when displaying the percents on the pie chart?  At present the pie chart is rounding to an integer (which was the original request), but the addition of this population level pie chart and the addition of the heredity level mutations means that we might need to display decimal digits (at least so we don't round values down to 0 which will confuse the user).  We show 2 decimal digits in the line chart hover and Frieda requested 1 decimal digit in the heredity level pie chart after looking at the mutation branch.  It doesn't need to be tackled now, but probably needs to be addressed during this pie chart work.

src/components/charts/pie-chart.tsx
```diff
@@ -53,17 +54,8 @@ export class PieChart extends BaseComponent<IProps, IState> {
...
-              case "#dfc39d":
-              case "#FEFF59":
-                labelColor = "#000000";
-                break;
-              default:
-                labelColor = "#ffffff";
-            }
-            return labelColor;
+            const backgroundColor = ctx.dataset.backgroundColor[ctx.dataIndex];
+            return colorIsLight(backgroundColor) ? "#000" : "#fff";
```
mklewandowski, line 58: We are probably going to need to have a general discussion about how to make this text visible when we have small, adjacent pie slices.  Might not need to be addressed now, but you can see in the image below that we're eventually going to need to come up with a solution to make the text easier to read in some cases:
image

src/models/spaces/populations/mouse-model/mouse-populations-model.ts
sfentress: This controls the line color of the populations graph. I think Michael went back on forth on it a few times, but these are the colors from https://hy7r7v.axshare.com/#id=8tunqt&p=connectedbio_-_specs__v2_7_3_ 

It's actually already in our css as `color-data-mousebrown-dark-rep`, and used in several other places such as the breeding pie charts.

mklewandowski: Good catch!

src/components/spaces/populations/populations-charts.tsx
sfentress: Agree with further discussion for later. The specs all show whole numbers, and this actually makes sense to me, but could understand thinking about using decimals. (Note, we shouldn't ever get 1% let alone >0.5% rounding down to 0, since the max carrying capacity of the environment is 50.)

src/components/charts/pie-chart.tsx
sfentress: Agreed.

src/components/spaces/populations/populations-charts.tsx
mklewandowski: Right.  And I should have been clear that I was conflating some of the issues that were seen in the heredity level pie charts when mutations were turned on.  It seems like this whole issue should probably be a separate story since it's tied to the work here in the population level and the mutation work in the heredity level.  The story should probably answer the following:
How do we want the pie chart numbers formatted?  
How do we want to handle visibility and display of the pie chart numbers?

sfentress: One thing (again, future discussion) that might help would be roll-over (plus click, four touch devices) of slices, which could show the exact value in the pop-up that appears (and maybe also raw numbers, not just percent). That would also potentially alleviate the crowded-labels issue. (I am not looking forward to thinking about dynamically generating external labels for small slices.)

src/models/spaces/populations/mouse-model/mouse-populations-model.ts
kswenson: Seems like a place where importing variables from SASS might make sense.

src/components/spaces/populations/populations-charts.tsx
mklewandowski: Yup, I groaned when I imagined creating and positioning the labels on the fly.  Maybe there's a more obvious solution?  Lots of folks must have this issue.

PR #170, opened 2020-05-01
Add mutations to heredity level breeding

mklewandowski:
This PR adds the ability to author mutation settings that work on the heredity level.  From the authoring page, users can enable a mutation checkbox, set the initial state of the checkbox, and specify the chance of mutation percentage.  Mutations appear in the nest pair litters when the breed button is pushed an litter offspring are created.

There are open issues regarding the percentages shown in the pie chart, but these are universal to all of the pie chart instances and are better dealt with in a separate PR.

### DIFF
--- a/src/authoring-schema.json
+++ b/src/authoring-schema.json
@@ -264,6 +264,18 @@
           "title": "Instructions (as markdown)",
           "type": "string"
         },
+        "enableStudentControlOfMutations": {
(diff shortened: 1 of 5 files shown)


mklewandowski:
I went ahead and moved the mutation code to `createGamete` - seems like a good decision.

PR #171, opened 2020-05-05
Pie charts slider

sfentress:
Updates the slider to control a pointer, which updates the values in the right-hand pie chart.

Note, this slider works slightly differently in each of four different states:

1. **No pie charts, All points**: Disabled slider
2. **No pie charts, Recent points**: As before: slider from 0-100% shifts chart to see from start to end
3. **Pie charts, All points**: Slider from 0-100% updates right chart from 0-100%
4. **Pie charts, Recent points**: Slider from 0-100% updates right chart from 0-100%, chart shifts so long as right-most point can be the pointer-point, and then remains static while pointer moves within the first 12 months of data.

### DIFF
--- a/src/components/charts/chart.sass
+++ b/src/components/charts/chart.sass
@@ -94,7 +94,11 @@
 
       .chart-container
         position: absolute
-        bottom: 59px
+        bottom: 75px
+
+      .line-chart-controls
+        position: absolute
+        bottom: 55px
 
     .top-chart
       display: flex
--- a/src/components/charts/chart.sass
+++ b/src/components/charts/chart.sass
@@ -100,6 +100,14 @@
         position: absolute
         bottom: 55px
 
+      .pointer-container
+        width: 352px
+        margin-left: 52px
+        height: 0px
+
+        .pointer
+          transform: translate(-50%, 6px)
+
     .top-chart
       display: flex
       justify-content: space-around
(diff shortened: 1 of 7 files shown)


sfentress:
Thanks for the detailed look and catches!

> If I pause the sim, switch to a different level (organism or heredity), and return to the population level, the slider is disabled and isn't re-enabled until I run the sim again.

Fixed.

> I notice that with no data points I can toggle the slider between the far left and right positions. This doesn't seem to matter, but potentially the slider should be disabled in this case.

Fixed.

> Is it potentially confusing that due to rounding we show "Current: x months" next to the pie chart for two positions of the slider? Instead of "x" and "x.5"? To further complicate this, if the user is scrolled to the far right of the line chart, the rounding can sometimes cause a mismatch between the label next to the pie chart and the line chart x axis label. See my attached pic where we show 36 months on the pie chart and 37 months on the line chart x axis.

We can show the team and see what they think. Personally I think it's ok because the 4.5th month is still the 4th month. I also think it would be really ugly for it to keep flashing between `x` and `x.5` every step, and I think using `x.0` would also be ugly. But happy to see what they think.

> On the heredity level, we used the text "light, medium, and dark" for the fur colors when the user hovers over the pie chart. On the population level, the addition of "brown" in the text means that the text "Medium brown" gets cut off when the user hovers over the pie chart.

Fixed.

> Do we expect that using the line chart legend to disable a chart series should also affect the pie chart? I 'm assuming no but wanted to point it out (since someone will turn a series off and ask why it isn't removed from the pie chart). See attached pic.

I see your point, but unsure what the expected behavior would be. I guess a white circle with only the one wedge, but that's not the way most pie charts work.

PR #180, opened 2021-01-22
Rename curriculum -> unit throughout

sfentress:
This fixes a small inconsistency where we have two names for the same thing and standardizes it on "unit," which matches what we use at CC more frequently.


mklewandowski (approved):
👍

PR #181, opened 2021-01-22
Make populations, organism models optional; Start to setup peas unit

sfentress:
This introduces the start of some branching based on whether the unit is `"mouse"` or `"pea"`. The largest change is that there is no populations or organism model in the Pea unit (yet), and so this PR adds support for optional models.

It also just barely starts making some UI distinctions between the units (the titles), and adds one small refactoring of mouse-specific files.

### DIFF
--- a/src/models/units.ts
+++ b/src/models/units.ts
@@ -1,0 +1,52 @@
+import { Unit } from "../authoring";
+
+interface UnitDefinition {
+  title: string;
+  populations: {
+    title: string;
+  };
+  organism: {
+    title: string;
+  };
+  breeding: {
+    title: string;
+  };
+  dna: {
+    title: string;
+  };
+}
+
+type Units = Record<Unit, UnitDefinition>;
+
+export const units: Units = {
+  mouse: {
+    title: "Deer Mice",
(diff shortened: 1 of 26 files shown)


src/models/units.ts
```diff
@@ -0,0 +1,52 @@
...
+      title: "Explore: DNA",
+    },
+  },
+  pea: {
+    title: "Peas",
+    populations: {
+      title: "Explore: Population",
+    },
+    organism: {
+      title: "Explore: Population",
```
mklewandowski, line 43: I assume this is meant to be a different string, but will probably be changed once the pea work is underway.

mklewandowski (approved):
👍 I think this looks good from a code review.  I was not, however, able to run locally.  I can no longer build and run on Windows since updating to Node 14 (I'm assuming that is the culprit).  I fail on what appears to be the install of canvas.  I tried downloading GTK and installing node-gyp, but so far I have not been able to install the dependencies.  I wonder if at some point we need to add to the readme if there are extra setup steps.  Regardless, this shouldn't hold up this PR.

PR #182, opened 2021-01-26
Species genetics

sfentress:
These defines the genetics (what the alleles are, what the breeding pairs are, what the phenotype is given a genotype) per species.

This unit and species definition will end up getting pretty large -- basically, any time there's going to be a difference between the pea unit and the mouse unit it will go here, to avoid code branches elsewhere in the app that needs to know about peas vs mice -- so it's possible this should get broken up.

Note there are still lots of class and model names that are mouse-specific (e.g. `BackpackMouse`), but refactoring those at this point would add a lot of noise to PRs. I will probably rename all one when things are more settled.


mklewandowski (approved):
👍 Changes look good.  I ran through the mouse version and everything worked as expected.

PR #183, opened 2021-01-27
Peas pie chart

sfentress:
This lets the species definitions define how the breeding pie charts are to be created.

Open to suggestions on improving the unit definitions as this gets bigger, since I'll be continuing with the plan that any difference between the units should be defined in those objects (this will need to eventually include the layout differences between mouse and plant breeding). The obvious thing of course will be to split them into different files when they get too big, but if there's something different you can think of let me know.

### DIFF
--- a/src/models/units.ts
+++ b/src/models/units.ts
@@ -1,5 +1,9 @@
 import { types } from "mobx-state-tree";
 import { Unit } from "../authoring";
+import { LegendItem } from "../components/spaces/breeding/breeding-data";
+import { PieChartData } from "../components/charts/pie-chart";
+// @ts-ignore
+import * as colors from "../components/colors.scss";
 
 export const UnitTypeEnum = types.enumeration("unit", ["mouse", "pea"]);
 
@@ -12,6 +16,8 @@ export interface UnitSpecies {
   phenotypeHeading: string;
   getPhenotypeLabel: (phenotype: string) => string;
   getGenotypeHTMLLabel: (genotype: string) => string;
+  chartTypes: Record<string, {legend: LegendItem[], title: string}>;
+  getChartData: (chartType: string, data: Record<string, number>) => PieChartData[];
 }
 
 interface UnitDefinition {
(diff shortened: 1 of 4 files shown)


src/models/units.ts
```diff
@@ -154,6 +197,31 @@ export const units: Units = {
...
+        chartTypes: {
+          phenotype: {legend: [
+            {label: "Smooth", color: colors.colorDataMouseBrownLightRep},
+            {label: "Wrinkled", color: colors.colorDataMouseBrownMediumRep}
+          ], title: "Pea Types" },
+          genotype: {legend: [
+            {label: "RR Peas", color: colors.colorDataMouseBrownLightRep},
+            {label: "Rr Peas", color: colors.colorDataMouseBrownMediumRep},
+            {label: "rR Peas", color: colors.colorDataMouseBrownMediumRep},
+            {label: "rr Peas", color: colors.colorDataMouseBrownDarkRep}
```
mklewandowski, line 209: I assume these will eventually be more "pea-specific" colors, but we don't have spec yet?

src/models/units.ts
```diff
@@ -154,6 +197,31 @@ export const units: Units = {
...
+        getChartData: (chartType, data) => {
+          let pieData = [];
+          if (chartType === "genotype") {
+            pieData = [{label: "RR", value: data.RR, color: colors.colorDataMouseBrownLightRep},
+                       {label: "Rr", value: data.Rr, color: colors.colorDataMouseBrownMediumRep},
+                       {label: "rR", value: data.rR, color: colors.colorDataMouseBrownMediumRep},
+                       {label: "rr", value: data.rr, color: colors.colorDataMouseBrownDarkRep}];
+          } else {
+            pieData = [{label: "Smooth", value: data.smooth, color: colors.colorDataMouseBrownLightRep},
+                       {label: "Wrinkled", value: data.wrinkled, color: colors.colorDataMouseBrownMediumRep}];
```
mklewandowski, line 221: Same with the colors in `getChartData` - we don't have a pea spec yet to set these colors, right?

src/models/units.ts
```diff
@@ -1,5 +1,10 @@
 import { types } from "mobx-state-tree";
 import { Unit } from "../authoring";
+import { LegendItem } from "../components/spaces/breeding/breeding-data";
+import { PieChartData } from "../components/charts/pie-chart";
+// @ts-ignore
+import * as colors from "../components/colors.scss";
```
mklewandowski, line 6: Just a note that we ended up using this method in CLUE and the AP for importing Sass variables:
https://mattferderer.com/use-sass-variables-in-typescript-and-javascript

I'm not saying that anything needs to change (especially since there are a handful of these imports in Cbio) - just mentioning it.

mklewandowski (approved):
👍

PR #184, opened 2021-02-10
Initial greenhouse UI

sfentress:
Hi @mklewandowski , this is a pretty large PR, which I apologize for. There wasn't an easy way to break up the work.

The vast majority of changes are repetitions on the same theme, and can probably just be skimmed in the Files Changed view:

1. Remove code that assumed something about the mouse layout, and make it a per-unit function that gets called from [unit.ts](https://github.com/concord-consortium/connected-bio-spaces/blob/initial-greenhouse-ui/src/models/units.ts)
2. Create a custom unit stylesheet, and add the unit name to the top-level container

For number 1 above, sometimes views call `units.ts` directly, and sometimes through the [backpack-mouse model](https://github.com/concord-consortium/connected-bio-spaces/compare/initial-greenhouse-ui#diff-1c9cfae9aa77200f5b0772e098300e550141b5db475fc86217b2d2b9c536449f), in particular for all the images (views used to call `organism.baseImage`, but now there is a wide variety of images they request). The annoying thing is that the logic for what image is shown when is totally different for mice vs peas, so we need a lot of custom image calls, many of which return the same image, but different sets return different "same images" for different species.


mklewandowski:
I don't know that any of these should slow down this PR, but here are a few things I noticed when running (some of these we saw yesterday so I know you're aware of them, but I'll mention them just in case they were worked on and we expect them to be fixed).  These could all probably be fixed in subsequent PRs.

Scroll offset:
image

Label vertical position:
image

Button label:
image

Legend color:
image

src/components/spaces/breeding/breeding-container.pea.sass
```diff
@@ -0,0 +1,15 @@
...
+    cursor: url('../../../assets/inspect-cursor-nests.png') 6 5, auto
+  &.breed
+    cursor: url('../../../assets/unit/pea/breeding_peas_cursor.svg') 6 3, auto
+  &.gametes
+    cursor: url('../../../assets/inspect-cursor-gametes.png') 6 5, auto
+
+.breeding-toolbar.pea
+  .toolbar-row.align-left
+    justify-content: center
+    margin-left: 0
```
mklewandowski, line 15: newline

src/components/spaces/breeding/breeding-view.pea.sass
```diff
@@ -0,0 +1,43 @@
...
+  .offspring
+    .litters-container
+      .litter
+        height: 88px
+        margin: 17px 0 -11px 0
+        background-image: url("../../../assets/unit/pea/pea_pod_transparent.png")
+        background-size: cover
+        animation: pea-pod-push-down 0.6s ease
+        .offspring-container
+          margin: $high-pea -3px;
```
mklewandowski, line 17: errant semi-colon

src/components/stacked-organism.sass
```diff
@@ -13,14 +13,16 @@
   &:hover .inspect-stack
     opacity: .75
 
+  *
+    pointer-events: none
+
```
mklewandowski, line 18: I didn't know about this wildcard syntax.  Neat.

mklewandowski (approved):
👍 This looks great!  I added a few comments with some visual/display issues (and very minor code tweaks), but I'm guessing most of this you're already aware of and can probably be tackled in future PRs if needed.

src/components/spaces/breeding/breeding-view.pea.sass
sfentress: :+1:

src/components/spaces/breeding/breeding-container.pea.sass
sfentress: :+1:

sfentress:
Fixed the linting errors, thanks. Yes, I didn't include the UI fixes in this PR, because basically, that's how it got so big: it's never done till it's done. I figured they'd go into the next.

src/components/stacked-organism.sass
sfentress: That's actually pure css (minus the sass nesting).

PR #189, opened 2021-02-17
Gametes toggle

sfentress:
This changes the logic of the gametes toggle switch. Previously, "showing gametes" was one of four different interactions
modes, which also included "breed" (selecting the nest), "select" and "inspect."

This was confusing, because "gametes" acted like "inspect", though it looked like its own independent toggle, and when users clicked to show gametes and then clicked on "inspect" they would leave "gametes" mode.

This disentangles the two, so that gametes being on or off is independent of the interaction mode.


dstrawberrygirl (approved):
I think this looks good!

PR #191, opened 2021-02-23
Reverse pea genotype label

sfentress:
This changes the standard genotype label so that the male allele is now show on the left.

This also swaps the gametes in the gamete-inspect panel such that the male gamete is on the left (matching the breeding panel).

Testable at

http://connected-bio-spaces.concord.org/branch/reverse-pea-genotype-label/index.html?%7B%22unit%22%3A%22pea%22%2C%22topBar%22%3Atrue%2C%22leftPanel%22%3Afalse%2C%22ui%22%3A%7B%22showBreedingSpace%22%3Atrue%2C%22investigationPanelSpace%22%3A%22breeding%22%7D%7D

(Breed some peas and turn on the gamete switch to show genotypes.)

### DIFF
--- a/src/components/inspect-panel.tsx
+++ b/src/components/inspect-panel.tsx
@@ -113,19 +113,28 @@ export class InspectPanel extends BaseComponent<IProps, IState> {
 
   private renderGametePanel = (mouse: BackpackMouseType) => {
     const species = speciesDef(mouse.species);
+    const unitBreeding = units[mouse.species].breeding;
     const arrowImage = "assets/unit/mouse/breeding/inspect/inspect-arrow.png";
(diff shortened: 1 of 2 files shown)


dstrawberrygirl (approved):
Certainly seems to work well! I was a bit confused at first glance by the flip in the label vs data on the chart in code, but in the application I see things render as I would expect so I think that's probably fine. Maybe worth adding a comment in that file to explain the flip?

sfentress:
It's basically a "flip" and a "reorder," but the change renders it as if the data itself got flipped.

Just wanted the chart legend labels to have "Rr" before "rR".

PR #202, opened 2021-06-29
remove pip install awscli and invalidate call in s3_deploy.sh

mklewandowski:
Remove pip install awscli and invalidate call in s3_deploy.sh to enable Travis builds (builds were failing).  Based on previous work in CLUE, GeoCode, and other projects:
https://github.com/concord-consortium/geocode/commit/ce24f88d550f9023c96be29a19c947bcf9c94afc

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ cache:
 # cf. https://stackoverflow.com/a/52387639
 install:
 - travis_retry gem install s3_website -v 3.4.0
-- travis_retry pip install awscli --upgrade --user
+#- travis_retry pip install awscli --upgrade --user
 - travis_retry npm ci
 - travis_retry cypress install
 script:
--- a/s3_deploy.sh
+++ b/s3_deploy.sh
@@ -56,4 +56,4 @@ s3_website push --site _site
 
 # explicit CloudFront invalidation to workaround s3_website gem invalidation bug
 # with origin path (https://github.com/laurilehmijoki/s3_website/issues/207).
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -10,7 +10,7 @@ cache:
 # cf. https://stackoverflow.com/a/52387639
 install:
 - travis_retry gem install s3_website -v 3.4.0
-- travis_retry pip install awscli --upgrade --user
+#- travis_retry pip install awscli --upgrade --user
```
kswenson, line 13: I seem to recall that @dougmartin figured out that if you install a newer version of python (3.6? 3.7?) before the `pip install awscli` line that everything works and you can keep the invalidation. The down side of removing the invalidation is that you may have to wait a bit from the end of the build before it can be tested. (It's supposed to be ~10 minutes but we've seen it appear to take longer.)

kswenson: I found this in the [Teacher Edition Tips Plugin](https://github.com/concord-consortium/teacher-edition-tips-plugin) repo by using GitHub search:
```
# the next command is to fix "SSLContext object is not available" error in awscli install
- travis_retry pyenv global 3.6
```

kswenson (changes requested):

kswenson:
I would recommend trying the `travis_retry pyenv global 3.6` before proceeding with removing the invalidation.

.travis.yml
mklewandowski: Yup, that solution works.  I'm going to close this PR and open a new one.

PR #203, opened 2021-06-29
update python version to fix pip error

mklewandowski:
Update the python version to the pip awscli install error on Travis which is causing our builds to fail.


kswenson (approved):
👍 LGTM

PR #206, opened 2021-07-01
Update gamete arrow positions in pea unit

mklewandowski:
This PR improves the positions of the gamete arrows in the peas unit.  I took the path of least resistance on this with the minimal number of changes, but if we ever add another unit, then we will probably want to rework this to be more versatile and use references to the divs that the arrows originate from and end at.

arrows


sfentress (approved):
Looks good to me, and I'm usually happy to live by the rule of not introducing complex generalizations until the third example.

PR #208, opened 2021-07-06
Add gamete case to pea parent inspect, improve assets

mklewandowski:
This PR updates the inspect images in the inspect panel in the Peas unit so that when the user has "show gamete" turned on, the inspect image appears with green background highlights instead of red background highlights.  This is in line with the current UI spec.

inspect


sfentress (approved):
Looks good :+1:

PR #209, opened 2021-07-15
Improve inspect/collect/gamete cursor states and highlight circles

mklewandowski:
This PR makes minor improvements to the cursor and organism highlight in relation to the show/hide gamete (based on discussion with Michael):
- collect cursor is shown when collect button is active (regardless of state of show/hide gametes)
- if inspect is active and we are hiding gametes, red inspect cursor is shown
- if inspect is active and we are showing gametes, organism highlight is green circle
- if inspect is active and we are hiding gametes, organism highlight is red circle
- if collect is active and either show or hide gametes is active, organism highlight is blue circle


sfentress:
Looks great :+1:

sfentress (approved):

